PR #8, opened 2020-01-10
CCC day fix up items

mark-kraemer:
https://trello.com/c/woYmIvtl/19-follow-up-to-ccc-days

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables.scss
@@ -47,6 +47,8 @@
   --font-primary: 'Roboto', sans-serif;
 
   // Text Ratios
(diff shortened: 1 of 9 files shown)


lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables.scss
```diff
@@ -47,6 +47,9 @@
   --font-primary: 'Roboto', sans-serif;
 
   // Text Ratios
+  // This is using a 16px base font-size. The scale ratio can be changed to
```
calebwoods: What ratio are we using?

luke-holloway: @mark-kraemer It is the minor third scale if you want to document that.

PR #14, opened 2020-06-01
Add Fujifilm Table Styles

jacobdmartin:
Imported the Fujifilm table styles to the index of the styleguide, and created a seperate table with radio button options.

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/card.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/card.scss
@@ -5,3 +5,10 @@
   width: 100%;
   background: var(--color-white);
 }
+
+.card--table {
+  display: flex;
+  flex-direction: column;
+
+  padding: 0;
+}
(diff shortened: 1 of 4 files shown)


example_rails6/Gemfile
```diff
@@ -48,3 +48,6 @@ end
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 gem 'rolemodel_rails', path: '..'
+
+gem 'react-rails'
+gem 'slim'
```
calebwoods, line 53: We don't want to commit the changes to the example Rails app as it is intended to be a base app with out any of our generators applied.

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/card.scss
```diff
@@ -5,3 +5,10 @@
   width: 100%;
   background: var(--color-white);
 }
+
+.card--table {
```
calebwoods, line 9: I thought we were going to rename this to `card--no-padding` per our discussion in the frontend team

jacobdmartin: You're right, my apologies. That comment must have gotten lost in my notes.

PR #18, opened 2020-06-22
Add generator for Heroku deploy setup

kylesmile:
Generator for basic Heroku deploy setup, as part of an effort to create instructions for setting up a Heroku server for a project. Since I only referenced a few LightningCAD projects, this may be incomplete.


lib/generators/rolemodel/heroku/templates/Procfile
```diff
@@ -0,0 +1,2 @@
+web: bundle exec puma -C config/puma.rb
+release: bin/heroku_release
```
calebwoods, line 2: I've seen most of our rails projects just have `bundle exec rails db:migrate` what else is the script getting us?

calebwoods: It looks the script handles the DB not existing, what's the way that happens?

kylesmile: `db:migrate` is probably sufficient. I think that script was adapted from something that was meant to decide whether it should run `schema:load` or `db:migrate`, but the consensus was that doing that was a bad idea. `app.json` covers running `schema:load` when creating a new server, anyway.

kylesmile: Removed the script in favor of calling `db:migrate` directly.

PR #25, opened 2020-08-04
Webpacker5

calebwoods:


lib/generators/rolemodel/webpacker/webpacker_generator.rb
```diff
@@ -6,14 +6,14 @@ class WebpackerGenerator < Rails::Generators::Base
...
-      node_version = '10.16.0'
+      node_version = '12.16.3'
       create_file ".node-version", node_version
       raise "Don't have node version #{node_version} installed. Fix it.\nRun: `nodenv install #{node_version}`" unless system 'node --version'
       raise "Don't have yarn installed. Fix it.\nRun: `npm install -g yarn`" unless system 'yarn --version'
     end
 
     def install_webpacker_with_react
-      gem 'webpacker'
+      gem 'webpacker', '~> 5.1'
```
calebwoods, line 16: This might be better with `gsub_file`. @timirwin would you be able to do that

timirwin: updated and should be ready to merge

PR #29, opened 2020-10-16
Css generator updates

mark-kraemer:


lib/generators/rolemodel/css/base/templates/app/views/styleguide/index.html.slim
```diff
@@ -1,202 +1,236 @@
...
+    title Style Guide
+    = csrf_meta_tags
+    = csp_meta_tag
+    = stylesheet_pack_tag 'stylesheets', 'data-turbolinks-track': 'reload'
+    meta name="viewport" content="width=device-width, initial-scale=1"
+  body class="#{'flex' if params[:sidebar]}"
+    - if params[:sidebar]
+      = render 'shared/sidebar'
+    - else
+      = render 'shared/top_nav'
```
calebwoods, line 13: Once someone chooses a layout, we might want them to actually use that in styleguide. Is there a simple way to make that migration path easier?

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/sidebar.scss
```diff
@@ -0,0 +1,37 @@
...
+  background: var(--color-white);
+  box-shadow: var(--shadow);
+
+  a {
+    text-decoration: none;
+  }
+}
+
+.sidebar-nav__item {
+  margin-bottom: 1rem;
```
calebwoods, line 19: This should use a `var`

calebwoods:
@mark-kraemer can you rebase this so we can get this integrated? I know the Shell Builder team is looking to use this pattern

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/sidebar.scss
scriswell: Ideally we would not use a margin here and instead use gap on a parent element.

scriswell:
@mark-kraemer @calebwoods this is an old PR. Is it still something we want merged in?

Jeremy-Walton:
This PR is not relevant anymore. CSS has been moved to https://github.com/RoleModel/rolemodel-design-system

PR #30, opened 2020-11-02
Update Heroku generator to enable force_ssl option in production

kylesmile:
Every app should be using SSL in production, especially on Heroku.

This generator change relies on Rails' default version of `config/environments/production.rb`, and assumes `force_ssl` is already set to `true` when the no commented-out `config.force_ssl = true` line exists. If that assumption isn't safe to make, I can make this more robust.


calebwoods (approved):
This should cover it as most apps will have this line in the config

PR #36, opened 2020-12-14
Add icons generator

kylesmile:
Adds a generator for a basic icon setup using material-icons. This doesn't support custom icons, but that could be added in a future PR (or possibly in this PR).

Most implementations of this pattern have put the `icon` helper method in `ApplicationHelper`. I opted to move this to its own helper file to reduce generator complexity and potential conflicts (I've tried this in an app and Rails does pick up the helper file).

### DIFF
--- a/lib/generators/rolemodel/css/icons/icons_generator.rb
+++ b/lib/generators/rolemodel/css/icons/icons_generator.rb
@@ -1,0 +1,20 @@
+module Rolemodel
+  module Css
+    class IconsGenerator < Rails::Generators::Base
+      source_root File.expand_path('templates', __dir__)
+
+      def add_material_icons
+        run 'yarn add material-icons'
+      end
+
+      def add_view_helper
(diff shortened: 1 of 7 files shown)


lib/generators/rolemodel/css/icons/icons_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Css
+    class IconsGenerator < Rails::Generators::Base
+      source_root File.expand_path('templates', __dir__)
+
+      def add_material_icons
+        run 'yarn add material-icons'
+      end
+
+      def add_view_helper
+        template 'app/helpers/icon_helper.rb'
```
timirwin, line 11: I believe you could use `copy_file` here too, since you aren't actually doing anything with the interpolating the template file itself.

timirwin (approved):
Should something also be added to the base README.md file?

PR #37, opened 2020-12-16
Add rolemodel:soft_destroyable generator

timirwin:
Add SoftDestroyable module and shared example for testing.

### DIFF
--- a/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
+++ b/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+module SoftDestroyable
+  extend ActiveSupport::Concern
+
+  included do
+    scope :kept, -> { where(deleted_at: nil) }
+    scope :only_deleted, -> { where.not(deleted_at: nil) }
+    define_model_callbacks :soft_destroy
+    define_model_callbacks :restore
+  end
+
+  module ClassMethods
+    def default_scope(scope = nil)
+      unless scope.nil? && !block_given?
+        raise "Default scopes should not be used with soft destroyable - in class #{name}"
+      end
+    end
+
+    def soft_destroy(timestamp = Time.zone.now)
+      kept.each { |o| o.soft_destroy(timestamp) }
+    end
+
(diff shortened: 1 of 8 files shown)


lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module SoftDestroyable
+  extend ActiveSupport::Concern
+
+  included do
+    scope :kept, -> { where(deleted_at: nil) }
+    scope :only_deleted, -> { where.not(deleted_at: nil) }
```
wesrich, line 8: Is it worth a `with_deleted` in case someone default_scopes and wants to get either one?

wesrich (approved):
One possible small suggestion, but it looks good to me!

lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
calebwoods: I wouldn't really recommend `default_scop` which was part of the why for this module

calebwoods (approved):

lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
wesrich: > I wouldn't really recommend `default_scope` which was part of the why for this module

Maybe not. And I'm not arguing for or against default_scope.

My thought was more, if someone should want an unscoped option, it might be nice to provide it. That may be due to a default_scope usage, or it may be just chaining scopes and wanting to definitively ensure they have it unscoped, but ... I guess it can always be added manually.

(For reference: https://github.com/jhawthorn/discard/blob/master/lib/discard/model.rb#L19 the Discarded gem provides a `with_discarded` scope to allow for the option, even though they also don't add a default_scope by default. I just think it's a nice to have, but wouldn't call it a blocker to not have one.)

timirwin: This implementation will `raise` if you try to define a `default_scope` in conjunction with SoftDestroyable.

PR #39, opened 2021-01-07
Add rolemodel:simple_form generator

timirwin:
Add simple_form generator for the gem, configuration, input files, and scaffold template


lib/generators/rolemodel/simple_form/templates/app/inputs/markdown_input.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class MarkdownInput < SimpleForm::Inputs::Base
```
calebwoods, line 3: Is there more JS that is needed for this input? Seems like might be a bit specific for the base generator.

lib/generators/rolemodel/simple_form/templates/config/initializers/simple_form.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  # You can wrap each item in a collection of radio/check boxes with a tag,
+  # defaulting to :span.
+  # config.item_wrapper_tag = :span
+
+  # You can define a class to use in all item wrappers. Defaulting to none.
+  # config.item_wrapper_class = nil
+
+  # How the label text should be generated altogether with the required text.
+  config.label_text = ->(label, required, _explicit_label) { "#{label} #{required}" }
```
calebwoods, line 116: I see this was left the default in Fujifilm https://github.com/RoleModel/fujifilm_cimba/blob/master/config/initializers/simple_form.rb
```suggestion
  # config.label_text = lambda { |label, required, explicit_label| "#{required} #{label}" }
```

Not sure what makes the most sense for our system default.

lib/generators/rolemodel/simple_form/templates/config/initializers/simple_form.rb
```diff
@@ -0,0 +1,182 @@
...
+  # config.custom_inputs_namespaces << "CustomInputs"
+
+  # Default priority for time_zone inputs.
+  # config.time_zone_priority = nil
+
+  # Default priority for country inputs.
+  # config.country_priority = nil
+
+  # When false, do not use translations for labels.
+  config.translate_labels = false
```
calebwoods, line 158: Seems odd to disable this

timirwin: Yes, I don't remember the details since Brent worked on that. There was a conflict between simple_form doing translation and probably LIT. I'll comment this out since it isn't likely that LIT would be used in the majority of cases anyway.

lib/generators/rolemodel/simple_form/templates/app/inputs/markdown_input.rb
timirwin: Probably don't need this. I'll remove it.

calebwoods (approved):

calebwoods:
Approved. Good baseline. I'm sure there is more we'd want to add as we go

PR #40, opened 2021-01-08
Update example6 app to Ruby 3.0 and Rails 6.1

timirwin:


example_rails6/babel.config.js
```diff
@@ -56,9 +56,7 @@ module.exports = function(api) {
       [
         '@babel/plugin-transform-runtime',
         {
-          helpers: false,
-          regenerator: true,
-          corejs: false
+          helpers: false
```
adamjsimon, line 59: Background on this change?

example_rails6/config/webpacker.yml
```diff
@@ -6,12 +6,11 @@ default: &default
   public_root_path: public
   public_output_path: packs
   cache_path: tmp/cache/webpacker
-  check_yarn_integrity: false
```
adamjsimon, line 9: Are we certain we want to keep this change? I know it is the new default out of Rails 6.1.1 but I am not certain we want to accept the change...

example_rails6/babel.config.js
timirwin: everything in the `example_rails6` folder is generated

example_rails6/config/webpacker.yml
timirwin: Perhaps we should add it into the webpacker generator if that is the preferred configuration, which is different from what is auto-generated? If we add it there, then it'll be part of our generator. If we modify the files here, it won't be in new apps and it won't be added by any generators either.

timirwin: @calebwoods Have any thoughts on this?

timirwin: Looks like the integrity check is `off` (https://github.com/rails/webpacker/issues/1568#issuecomment-682495933) by default now because it was causing confusion. If we want it `on` in development, sounds like that change should be in development.rb instead of webpacker.yml (https://github.com/rails/webpacker/issues/1568#issuecomment-459995382). As mentioned, I think that'd be part of the webpacker generator.

calebwoods: I would agree we should make the change in the webpacker generator if we want to change the default. The example apps are meant to be vanilla so we can test running against something simple

calebwoods (approved):

PR #41, opened 2021-01-11
Update Devise generator with translated content

timirwin:


lib/generators/rolemodel/saas/devise/templates/app/views/devise/registrations/new.html.slim
```diff
@@ -1,15 +1,15 @@
 .flex.flex-col.items-center
-  h2.margin-top-lg Sign up
+  h2.margin-top-lg= t(".sign_up", default: "Sign up")
   .card.container--xs.margin-top-md
     = simple_form_for(resource, as: resource_name, url: registration_path(resource_name)) do |f|
       = f.input :first_name, required: true, autofocus: true
       = f.input :last_name, required: true
       = f.simple_fields_for :organization do |organization_field|
-        = organization_field.input :name, label: 'Organization name'
+        = organization_field.input :name, label: t(".organization_name", default: "Organization name")
```
calebwoods, line 8: Shouldn't the normal AR translations handle this? Do we really need to set the label manually?

lib/generators/rolemodel/saas/devise/templates/app/views/devise/sessions/new.html.slim
```diff
@@ -8,8 +8,8 @@
         .form__group
           = f.label :remember_me
             = f.check_box :remember_me
-            span.margin-left-xs Remember me
+            span.margin-left-xs= t("activerecord.attributes.user.remember_me", default: "Remember me")
```
calebwoods, line 11: This could probably be replace with a `f.input` since we are using SimpleForm right?

timirwin: Sure, we can change that. Could you approve the simple_form PR (https://github.com/RoleModel/rolemodel_rails/pull/39)? 😉

lib/generators/rolemodel/saas/devise/templates/app/views/devise/registrations/new.html.slim
timirwin: I'll give it a look. I feel like we need to override it because it was defaulting to just "Name" which isn't very descriptive in this context.

lib/generators/rolemodel/saas/devise/templates/app/views/devise/sessions/new.html.slim
timirwin: fixed

lib/generators/rolemodel/saas/devise/templates/app/views/devise/registrations/new.html.slim
timirwin: So, it does just show "Name" as expected, but I changed it to have organization name be in the devise.en.yml to be translated as "Organization name".

calebwoods (approved):

PR #42, opened 2021-01-13
Add admin/users controller for user edit and reinvite

timirwin:


lib/generators/rolemodel/saas/devise/templates/app/views/admin/users/index.html.slim
```diff
@@ -0,0 +1,27 @@
+header.header--index-action
```
calebwoods, line 1: Does this match the default HTML we have for some of the other Devise generators? Probably something we'll want to iterate on further as we add more updates/better naming to RoleModel CSS

lib/generators/rolemodel/saas/devise/templates/app/views/admin/users/index.html.slim
```diff
@@ -0,0 +1,27 @@
...
+  h1 Users
+  - if Devise.mappings[:user].try(:invitable?)
+    = link_to 'Add User', new_invitation_path(:user), class: 'btn btn--primary'
+  = link_to 'Sign out', destroy_session_path(current_user), method: :delete
+
+.table__wrapper
+  table.table data-controller="clickable-rows"
```
calebwoods, line 8: Clickable Rows is a Stimulus controller from BGEA right? I don't think it is something we've brought into the generator yet, or have we?

timirwin: good catch - yeah, that wasn't intended to get in there

calebwoods:
@timirwin what would it take to get this merged in?

timirwin:
> @timirwin what would it take to get this merged in?

I was expecting an approval for the PR, but maybe you implied that. I'll go ahead and merge.

PR #45, opened 2021-01-21
Add panels and modals

timirwin:


timirwin:
@mark-kraemer This is at least a good start. Feel free to correct my mistakes.

lib/generators/rolemodel/modals/templates/app/javascript/stylesheets/components/panel.scss
```diff
@@ -0,0 +1,23 @@
...
+  position: absolute;
+  right: calc(-1 * var(--panel-width)); // this pushes the panel off the sceen to the right
+  background: white;
+  transition: right var(--panel-transition-speed);
+
+  overflow-y: scroll;
+  padding-bottom: var(--space-lg);
+}
+
+
```
mark-kraemer, line 14:
```suggestion
```

lib/generators/rolemodel/css/base/base_generator.rb
```diff
@@ -12,22 +12,16 @@ def add_slim
...
       def add_simple_form
-        gem 'simple_form'
-        run_bundle
-
-        generate 'simple_form:install'
-
-        gsub_file 'config/initializers/simple_form.rb', 'config.wrappers :default, class: :input', 'config.wrappers :default, class: :form__group'
-        gsub_file 'config/initializers/simple_form.rb', '# config.label_class = nil', 'config.label_class = :form__label'
-        gsub_file 'config/initializers/simple_form.rb', '# config.input_class = nil', 'config.input_class = :form__input'
+        generate 'rolemodel:simple_form'
```
calebwoods, line 15: If we now have this in separate generator, do we need to run it again here?

calebwoods: Should do a conditional like we are using for the icon generator?

lib/generators/rolemodel/modals/modals_generator.rb
```diff
@@ -0,0 +1,43 @@
...
+      generate 'rolemodel:css:icons'
+    end
+
+    def add_ui_components
+      copy_file 'app/helpers/dynamic_link_helper.rb'
+      addendum = File.read([source_paths.last, '/application_addendum.js'].join)
+      append_to_file 'app/javascript/packs/application.js', addendum
+      copy_file 'app/javascript/packs/rolemodel-custom-confirm.js'
+      copy_file 'app/javascript/packs/rolemodel-modal.js'
+      copy_file 'app/javascript/packs/rolemodel-panel.js'
```
calebwoods, line 16: Doesn't `copy_file` need a destination path? Also do we really want to have all these files in the root of `packs` if so they will get emitted as JS bundles by webpack (I think). We don't ever intend to use them directly so it would be it would make sense for them to go in a sub folder

lib/generators/rolemodel/modals/templates/app/javascript/packs/rolemodel-modal.js
```diff
@@ -0,0 +1,54 @@
...
+      this.closeModalElement.addEventListener('click', (e) => {
+        e.preventDefault()
+        this.modal.classList.remove('modal--active')
+        setTimeout(() => { this.modalBody.innerHTML = ''}, 1000)
+      })
+    }
+  }
+
+  displayError() {
+    this.modalBody.innerHTML = '<div class="page-load__error"><span class="material-icons">error</span><p>Uh oh! Something broke. Try reloading...</p></div>'
```
calebwoods, line 52: Would it be possible to declare this as a template (JS template literal) or partial somewhere? Just seeing that this is repeated at least once

lib/generators/rolemodel/modals/templates/app/javascript/packs/rolemodel-panel.js
```diff
@@ -0,0 +1,52 @@
...
+        this.displayPanelError()
+      })
+    })
+  }
+
+  enablePanelClose() {
+    if (this.closePanelElement) {
+      this.closePanelElement.addEventListener('click', () => {
+        this.panel.classList.remove('panel--active')
+        setTimeout(() => { this.panelContent.innerHTML = '' }, 1000) // just wait until the animation finishes (400ms)
```
calebwoods, line 44: Should we be calculating the setTimeout length based on the animation timing? And then just default to a second?

lib/generators/rolemodel/modals/templates/app/javascript/stylesheets/components/modal.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.modal__backdrop {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 11;
+  background: var(--color-black);
+	visibility: hidden;
```
calebwoods, line 27: Indent

lib/generators/rolemodel/modals/templates/app/javascript/stylesheets/components/modal.scss
```diff
@@ -0,0 +1,66 @@
...
+  opacity: 0;
+  transition: all var(--modal-transition-speed);
+}
+
+.modal__content {
+  max-height: 500px;
+  width: 564px;
```
calebwoods, line 34: This width seems pretty arbitrary. Should this be coming from a variable?

lib/generators/rolemodel/modals/templates/app/views/layouts/full_screen.html.slim
```diff
@@ -0,0 +1,15 @@
...
+    title App Name
+    = csrf_meta_tags
+    = csp_meta_tag
+    = stylesheet_pack_tag 'stylesheets', 'data-turbolinks-track': 'reload'
+    = javascript_pack_tag 'application', 'data-turbolinks-track': 'reload'
+
+  body
+    .container
+      = render 'shared/flash'
+      = render 'shared/breadcrumbs'
```
calebwoods, line 13: breadcrumbs won't exists for every project

lib/generators/rolemodel/modals/templates/app/views/shared/_custom_confirm.html.slim
```diff
@@ -0,0 +1,8 @@
+.modal#custom-confirm.flex.items-center.justify-center
```
calebwoods, line 1: I'm thinking we could actually drop the "custom" prefix on this naming. Couldn't we just call this RolemodelConfirm? We were using a custom to distinguish from the Rails built-in, but now we are fully overriding

lib/generators/rolemodel/modals/modals_generator.rb
mark-kraemer: If the file path the is the same as the destination, there is no needed to add a second argument.

lib/generators/rolemodel/modals/templates/app/javascript/packs/rolemodel-modal.js
mark-kraemer: Done

lib/generators/rolemodel/modals/templates/app/javascript/stylesheets/components/modal.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.modal__backdrop {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 11;
+  background: var(--color-black);
```
calebwoods, line 26: This needs to be refactored when we merge color theory

lib/generators/rolemodel/modals/templates/app/javascript/stylesheets/components/panel.scss
```diff
@@ -0,0 +1,22 @@
+.panel {
+  height: 100vh;
+  width: var(--panel-width);
+  z-index: 100;
+  position: absolute;
+  right: calc(-1 * var(--panel-width)); // this pushes the panel off the sceen to the right
+  background: white;
```
calebwoods, line 7: Need variables

lib/generators/rolemodel/modals/templates/app/helpers/dynamic_link_helper.rb
```diff
@@ -0,0 +1,9 @@
+module DynamicLinkHelper
+  def panel_link_to(name, url, classes = nil)
```
calebwoods, line 2: Want to make this signature the same as link_to

PR #60, opened 2021-07-09
Improve webpacker compiling behavior in test mode

kylesmile:
When Webpacker is configured to automatically compile assets, it makes a potentially slow check for changed files every time something wants to look up a Webpack-managed resource (so _every_ call on a page to thing like `stylesheet_pack_tag`, `javascript_pack_tag`, or `image_pack_tag` will make this check).

This can be fine in development mode (particularly with the webpack dev server running, because then the check is skipped), but it can significantly slow down tests. On Dock/Railing Designer, we had RSpec test time go from ~1min 30sec to ~20sec with these changes.


lib/generators/rolemodel/webpacker/webpacker_generator.rb
```diff
@@ -57,6 +57,11 @@ def add_jest_config
       copy_file 'FilePathMock.js', 'spec/javascript/support/FilePathMock.js'
     end
 
+    def add_rspec_helper
+      say 'Add RSpec helper'
+      copy_file 'spec/support/webpacker.rb'
```
kylesmile, line 62: I wasn't 100% sure if this helper should be in the Webpacker generator or the Testing generator, since it's related to testing when Webpacker is in use. Thoughts?

calebwoods: Webpacker is added by default. So I think this actually makes sense to be in the testing generator

calebwoods (approved):
Just moving the testing generator makes sense. Would be good to try testing this on a project like Ninja Master see what the improvement is

kylesmile:
For Ninja Master, I'm getting around 12min 3sec without this change and around 10min 33sec with the change when running RSpec locally. So a relatively smaller change for Ninja Master than for Dock Designer, but still seems to be an improvement.

calebwoods (approved):

calebwoods (approved):

PR #63, opened 2021-08-03
Grid utilities

Jeremy-Walton:
## Why?

In order to provide simpler ways to leverage flex and its layout capabilities, we want to add a column system inspired by https://bulma.io/documentation/columns/ with spacing, offsets, semantic names

## What Changed

* [X] Add column styles
* [X] Add Semantic spacing
* [X] Add Semantic Offsets
* [X] Add 12 Point Sizing
* [X] Add 12 Point Offsets
* [X] Add Variable Gap support
* [X] Add force mobile flex support

## Screenshots

Screen Shot 2021-08-03 at 11 43 25 AM
Screen Shot 2021-08-03 at 11 43 29 AM
Screen Shot 2021-08-03 at 11 43 39 AM
Screen Shot 2021-08-03 at 11 44 36 AM
Screen Shot 2021-08-03 at 11 44 40 AM

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
@@ -1,0 +1,88 @@
+.columns {
+  display: flex;
+}
+
+.column {
+  display: block;
+  flex-basis: 0;
+  flex-grow: 1;
+  flex-shrink: 1;
+  padding: var(--space-sm);
+
+  //  Sizing
+
+  &.is-full {
+    flex: none;
+    width: 100%;
+  }
+
+  &.is-half {
+    flex: none;
+    width: calc(100% * 1/2);
+  }
+
+  &.is-offset-one-half {
+    margin-left: calc(100% * 1/2);
+  }
+
+  @function pluralize($string, $number) {
+    @if $number == "1" {
+      @return $string;
+    } @else {
+      @return "#{$string}s";
+    }
+  }
+
+  // Fifths Sizing and Offset
+  $numbers: ("four": "4", "three": "3", "two": "2", "one": "1");
+  @each $name, $number in $numbers {
+    &.is-#{$name}-#{pluralize("fifth", $number)} {
+      flex: none;
+      width: calc(100% * #{$number}/5);
+    }
+
+    &.is-offset-#{$name}-#{pluralize("fifth", $number)} {
+      margin-left: calc(100% * #{$number}/5);
+    }
+  }
+
+  // Quarters Sizing and Offset
+  $numbers: ("three": "3", "one": "1");
+  @each $name, $number in $numbers {
+    &.is-#{$name}-#{pluralize("quarter", $number)} {
+      flex: none;
+      width: calc(100% * #{$number}/4);
+    }
+
+    &.is-offset-#{$name}-#{pluralize("quarter", $number)} {
+      margin-left: calc(100% * #{$number}/4);
+    }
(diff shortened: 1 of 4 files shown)


jdmcleod (approved):

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
```diff
@@ -0,0 +1,444 @@
...
+// region Fifths Sizing and Override
+.column.is-four-fifths {
+  flex: none;
+  @media screen and (min-width: $breakpoint-sm) {
+    width: var(--four-fifths);
+  }
+}
+
+.column.is-offset-four-fifths {
+  @media screen and (min-width: $breakpoint-sm) {
```
Jeremy-Walton, line 183: I can move the media query to be above all the offsets and sizings

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// Variable Gap Sizing
+.columns.is-variable {
+  --columnGap: var(--space-sm);
+  margin-left: calc(-1 * var(--columnGap));
+  margin-right: calc(-1 * var(--columnGap));
+
+  // TODO: better name? These could be confused with the column sizings
+  // E.G. `.columns.is-variable.is-1` is close to `.column.is-1`
```
Jeremy-Walton, line 69: Any ideas on a better pattern for this?

Jeremy-Walton:
Note: this doesn't use `grid-gap` which I know @calebwoods had mentioned as a possibility. I don't think we need it yet to accomplish good things with this system.

Jeremy-Walton:
Also, the naming convention differs from our standard BEM conventions. I think for this system it makes sense to have Semantic, readable names. `.columns.columns--variable.columns--gap1` compared to `.columns.is-variable.is-1(or .is-1-gap)`. `.column.column--one-third.column--offset-one-third` compared to `.column.is-one-third.is-offset-one-third`.

Talking with @scriswell I think we agree that the different convention makes sense in this context.

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
Jeremy-Walton: `.is-gap-one`?

Jeremy-Walton:
Any thoughts on how best to document how to use these?

connortorrell (approved):

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/columns.scss
mark-kraemer: Yeah, I think being more explicit with the class would make it easier to read and pick up for new devs. This would also match with current utility class methods

```css
.is-gap-none
.is-gap-xxs
.is-gap-xs
.is-gap-sm
etc...
```

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/styleguide.scss
```diff
@@ -31,3 +31,19 @@
 
   background: var(--color-primary);
 }
+
+.columns-example {
```
Jeremy-Walton, line 35: This isn't used anymore and can go away

scriswell (approved):

PR #64, opened 2021-08-03
Turbo setup

Jeremy-Walton:
## Why?

Since we are starting to adopt Turbo usage more often, we want to simplify the setup and add fixes and other changes we have found to be helpful.

## What Changed

* [X] Update example_rails6 mimemagic gem version
* [X] Add Turbo Generator to replace turbolinks
* [X] Update Modals generator to require turbo first
* [X] Change Modals generator event listener to listen for `turbo:load` instead of `turbolinks:load`
* [X] Update All generator to run turbo generator

### DIFF
--- a/example_rails6/Gemfile.lock
+++ b/example_rails6/Gemfile.lock
@@ -102,7 +102,9 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (1.0.0)
-    mimemagic (0.3.5)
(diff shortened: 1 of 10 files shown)


connortorrell (approved):

timirwin (approved):

Jeremy-Walton:
This is no longer relevant. This was dependent on RailsUJS and that is being deprecated in Rails 7. This installation is partly baked into Rails 7 anyway.

PR #65, opened 2021-08-03
Add Custom Icons to Icon Generator

jdmcleod:
# Add custom Icons to Icon Generator

The desire here is to be able to reference custom icons in addition to material icons with the `#icon` helper method. Currently, material icons can be rendered with `= icon(<name>)`. The goal is to have the ability to similarly render custom icons that live inside `app/javascript/images/icons`

This ports over two webpack loaders from sayfa to optimize svg files and allow the ability to override the color. One of the goals is to have the icons behave similar to material icons in that, if they don't have a specified color, they will inherit the font color of the parent css element.

### Misc
- [X] Fixed an issue with `--neutral-layer-app-text` not being included in dark mode
- [X] Set config.boolean_style to `inline` in simple_form generator
- [X] Modify css base generator to automatically run webpacker generator if not installed (follow pattern in modal generator) 

### Modifications to icon generator
- [X] Inject code into `application.js` to require files in `app/javascript/images` so that webpack compiles them
- [X] Use two custom webpack loaders from sayfa to normalize svgs and make them behave more like icons
- [X] Modify webpack file loader to exclude files in `/images/icons` directory to avoid processing them in two different loaders
- [X] Add "Icons" Section to styleguide 
- [X] Modify `icon_helper.rb` to grab custom icons from webpack
- [X] Add `svgo` npm package
- [x] Allow custom icons to inherit font color of parent class
- [x] Allow custom classes to change size of custom icon

## Screenshots
Screen Shot 2021-08-03 at 7 07 22 PM


lib/generators/rolemodel/css/base/base_generator.rb
```diff
@@ -6,6 +6,10 @@ class BaseGenerator < Rails::Generators::Base
       include Rolemodel::BundlerHelpers
       source_root File.expand_path('templates', __dir__)
 
+      def install_webpack
+        generate 'rolemodel:webpacker' unless Gem.loaded_specs.has_key?('react-rails')
```
timirwin, line 10: Checking if the `react-rails` gem is included before generating `webpacker`. Could you check for the `webpacker` gem instead?

lib/generators/rolemodel/css/icons/icons_generator.rb
```diff
@@ -3,18 +3,120 @@ module Css
...
+      end
+
       def add_css
         copy_file 'app/javascript/stylesheets/components/icon.scss'
         append_file 'app/javascript/packs/stylesheets.scss', "@import 'stylesheets/components/icon';"
       end
+
+      def add_image_directories
+        run 'mkdir app/javascript/images'
+        run 'mkdir app/javascript/images/icons'
```
timirwin, line 32: Should we use `-p` and simplify to a single line?

```suggestion
        run 'mkdir -p app/javascript/images/icons'
```

Or use Thor?
```suggestion
        empty_directory 'app/javascript/images/icons'
```
Or should we just use Ruby?
```suggestion
        FileUtils.mkdir_p Rails.root.join('app/javascript/images/icons')
```

timirwin (approved):
A couple of things that you might want to review

lib/generators/rolemodel/css/base/base_generator.rb
jdmcleod: I believe I started out by checking for `webpacker`, but it did not work, I think because it is installed even before the webpacker generator is run (not sure why). I found that checking for` react-rails` to be present did the right thing, but agree with you that this could be improved.

lib/generators/rolemodel/css/icons/icons_generator.rb
jdmcleod: I like the third one!

lib/generators/rolemodel/css/icons/templates/config/webpack/loaders/svg-color-override-loader.js
```diff
@@ -0,0 +1,43 @@
+const { optimize, extendDefaultPlugins } = require('svgo')
+
+module.exports = function(content) {
+
+  // TODO: We should add some documentation here explaining what this config doing.
```
kylesmile, line 5: I'll push a commit to for this.

kylesmile (commented):
The example icon doesn't have any color attributes set, which is why inheriting the font color wasn't working. I'll push a commit to handle that case.

Generally seems fine to me, though.

lib/generators/rolemodel/css/icons/icons_generator.rb
```diff
@@ -3,18 +3,119 @@ module Css
...
+          slim
+        end
+      end
+
+      def require_images
+        inject_into_file 'app/javascript/packs/application.js', after: "ActiveStorage.start()\n" do
+          <<~'JS'
+
+            // require custom icons
+            require.context('images')
```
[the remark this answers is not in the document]
jdmcleod: I think it is fine to let it be all images. The icon loaders should only pick up what is inside `/icons` so I don't think it hurts. What do you think?

lib/generators/rolemodel/css/icons/templates/app/helpers/icon_helper.rb
```diff
@@ -1,5 +1,42 @@
 module IconHelper
-  def icon(name, classes: nil, color: nil, hover_text: nil)
-    tag.span(name, class: "material-icons #{classes}", style: "color: var(--color-#{color})", title: hover_text)
+  def icon(name, classes: nil, color: nil, hover_text: name)
+    contents = custom_icon(name) || name
+
+    tag.span(contents, class: icon_classes(classes, custom_icon(name)), style: "color: var(--color-#{color})", title: hover_text)
```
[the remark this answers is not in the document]
jdmcleod: I don't know of a way other than asking webpack if it exists with `Webpacker.manifest.lookup`. The best I can think of is to extract that bit of logic out.

lib/generators/rolemodel/css/icons/icons_generator.rb
```diff
@@ -100,7 +100,7 @@ def modify_stylguide_with_examples
                 background: var(--color-neutral-50);
                 border-radius: 3px;
```
calebwoods, line 101: This should probably use our radius variable

lib/generators/rolemodel/css/base/base_generator.rb
```diff
@@ -6,6 +6,10 @@ class BaseGenerator < Rails::Generators::Base
       include Rolemodel::BundlerHelpers
       source_root File.expand_path('templates', __dir__)
 
+      def install_webpack
+        generate 'rolemodel:webpacker'
```
calebwoods, line 10: Not sure I'd add this here. Is something now dependent on that?

lib/generators/rolemodel/css/icons/README.md
```diff
@@ -1,8 +1,14 @@
...
 ## What you get
 
-* material-icons and the necessary CSS setup
-* A view helper for icons (`icon(name)`)
+### Material Icons
+* Allows use of material-icons with the necessary CSS setup. 
+* Adds a icon helper to render them `icon(<material-icon-name>)`
+
+### Custom Icons
+* Adds ability to use custom icons by putting images into `app/javascript/images` and referencing them with icon helper `icon(<custom-icon-name>)`
```
calebwoods, line 12:
```suggestion
* Adds ability to use custom icons by putting images into `app/javascript/images/icons` and referencing them with icon helper `icon(<custom-icon-name>)`
```

lib/generators/rolemodel/css/icons/USAGE
```diff
@@ -7,6 +7,9 @@ Example:
     This will create:
         app/helpers/icon_helper.rb
         app/javascript/stylesheets/components/icon.scss
+        app/javascript/images
```
calebwoods, line 10:
```suggestion
        app/javascript/images/icons
```

calebwoods (approved):
A couple of minor things to clean up, but otherwise looks good

lib/generators/rolemodel/css/base/base_generator.rb
jdmcleod: I figured the webpacker generator was required for css base since all the styles are served through webpacker. This way, icons depends on css, and css depends on webpacker. Do you see a scenario where the css generator would be run without running the webpacker one?

calebwoods: I don't see a case where it won't be run. It just probably has already been run by the time we get to this point. Don't really need/want to run it again in all cases

lib/generators/rolemodel/css/icons/icons_generator.rb
```diff
@@ -3,18 +3,119 @@ module Css
...
+              },
+              { after: 'file' }
+            )
+  
+            environment.loaders.get('file').exclude = /\/icons\/.*.svg$/
+          JS
+        end
+      end
+
+      def modify_stylguide_with_examples
```
jdmcleod, line 54: typo

PR #76, opened 2021-11-21
Fix "Fixed width flex items" overlapping container when combined with gap

Jeremy-Walton:
## Why?

If you set a fixed width on all items of a flex container that is using layout-row and gap, they overlap the container due to the gap adding space.

## What Changed

* [X] Change layout-row to apply flex-basis for the width utilities so it still obeys the flex container.

## Screenshots

Old:
Screen Shot 2021-11-20 at 11 03 54 PM

New:
Screen Shot 2021-11-20 at 11 01 02 PM

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
@@ -54,10 +54,19 @@
   & > * {
     flex: 1;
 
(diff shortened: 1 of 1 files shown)


scriswell (approved):
@Jeremy-Walton This looks good, is it up to date?

Jeremy-Walton:
This was included in the massive design update so I think this PR is no longer relevant. These changes are already in master.

PR #81, opened 2022-01-14
[TR #66] restructure-card-file

zoopmaster:
## Task

[TR #59](https://trello.com/c/HX4dt7Vc/59-e48-use-box-shadow-instead-of-border)
[TR #63](https://trello.com/c/ptvCQBQC/63-e48-remove-card-rounded)
[TR #65](https://trello.com/c/zLVmxAZb/65-e48-h-elements-should-reset-using-margin-0-instead-of-unset)
[TR #66](https://trello.com/c/lQU5vlKl/66-e48-structure-the-file-using-a-card)
[TR #78](https://trello.com/c/ucq5uZNW/78-import-line-height-variables-from-motz)
[TR #79](https://trello.com/c/0ygsOFiP/79-import-border-variables-from-motz)

## Why?

Implement a number of structure and styling simplifications to base card styles for easier and more consistent development and styling.

## What Changed

What changed in this PR?

* [ ] Adjusted `border-radius` tokens
* [ ] Added `line-height` tokens
* [ ] Refactored default card structure styling
* [ ] Begin replacing `borders` with `box-shadows` tokens
* [ ] Added simple documentation to `.card.scss`

## Screenshots

New border using `box-shadow` and adjusted `border-radius` scale
image

PR #83, opened 2022-02-08
[TR #81] Add separate concepts for layout row and flexible layout area

Jeremy-Walton:
## Task

[TR #81](https://trello.com/c/r7I7rdwk)

## Why?

The layout-row utility was overloaded and doing more than it should have.

It needs to be simplified and a new concept `flexible-layout-area` needs to be added to clearly separate the concepts and add documentation.

## What Changed

* [X] Simplify layout-row
* [X] Add flexible-layout-area
* [X] Add documentation to both
* [X] Move layout utilities to the bottom of the file.
* [X] Fix gap-sm helper (was pointing at xxs spacing)

## Screenshots

Screen Shot 2022-02-08 at 10 23 42 AM
Screen Shot 2022-02-08 at 10 23 48 AM
Screen Shot 2022-02-08 at 10 24 15 AM
Screen Shot 2022-02-08 at 10 24 20 AM

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
@@ -24,70 +24,35 @@
 
 .flex-wrap { flex-wrap: wrap; }
 
+.flex-no-wrap { flex-wrap: nowrap; }
+
 .flex-grow-1 { flex-grow: 1; }
 
-// Layout properties
-// E.G:
-// .flex.layout-row.gap-md
-//   .two-thirds-width.card Item
-//   .card Item
-//   .card Item
-
-// Normal width
-// |      Item     | | Item | | Item |
-// Mobile
-// | Item |
-// | Item |
-// | Item |
-
-// E.G:
-// .flex.layout-row.gap-md.flex-wrap
-//   .two-thirds-width.card Item
-//   .card Item
-//   .card Item
-
-// Normal width
-// |      Item     | | Item | | Item |
-// Mobile
-// |   Item   | | Item |
-// |        Item       |
-
-.layout-row {
-  & > * {
-    flex: 1;
-
-    &.four-fifths-width { flex-basis: calc(100% * 4/5); }
-    &.three-fifths-width { flex-basis: calc(100% * 3/5); }
-    &.two-fifths-width { flex-basis: calc(100% * 2/5); }
-    &.one-fifth-width { flex-basis: calc(100% * 1/5); }
-
-    &.three-quarters-width { flex-basis: calc(100% * 3/4); }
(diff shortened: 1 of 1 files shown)


lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
```diff
@@ -24,70 +24,35 @@
...
 // Gap Properties
-.gap-xxs { gap: var(--space-xxs); }
-.gap-xs { gap: var(--space-xs); }
-.gap-sm { gap: var(--space-xxs); }
-.gap-md { gap: var(--space-md); }
-.gap-lg { gap: var(--space-lg); }
-.gap-xl { gap: var(--space-xl); }
+.gap-xxs {
+  gap: var(--space-xxs);
+  --gap: var(--space-xxs);
```
scriswell, line 34: @Jeremy-Walton Why do we need to add the custom properties within the classes here?

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
```diff
@@ -286,111 +233,154 @@
...
+.margin-left-md { margin-left: var(--space-md); }
+.margin-left-sm { margin-left: var(--space-sm); }
+.margin-left-xs { margin-left: var(--space-xs); }
+.margin-left-none { margin-left: 0; }
 
-.margin-bottom-sm {
-  margin-bottom: var(--space-sm);
-}
+// Elevations
+.elevation-1, .elevation-1.card { box-shadow: var(--elevation-1); }
```
scriswell, line 271: @Jeremy-Walton Any way we could just get this done with only the `.elevation-[#]` classes and drop `.elevation-[#].card`?

scriswell (commented):
@Jeremy-Walton I have a few questions. I'd also like to pair with you on naming and share a few ideas with you.

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/utilities.scss
Jeremy-Walton: The reason for that is so that when something under an element with a gap set, it can have access to specifically which gap is set (via the --gap variable). The specific use case is when using `flexible-layout-area` with a gap set and using the width utilities e.g. `one-fifth-width` under it. Under the `flexible-layout-area`, those width utils use `flex-basis` to calculate width which does not take into account any gap set, so I had to make a way that they could include the gap when calculating the width. e.g. `.one-fifth-width { flex-basis: calc(calc(100% * 1/5) - var(--gap)); }`.

If I didn't set the `--gap`, I wouldn't know which spacing variable to subtract out here

scriswell (approved):

PR #84, opened 2022-02-08
Add production default mailer config

mark-kraemer:
## Why?

Pull in our default production mailer config. This also includes a fix for premailer to get the right asset host set

## What Changed

What changed in this PR?

* [x] Fix premailer asset host issue with image tag for logos 
* [x] Add in default sendgrid config to production
* [x] Add staging whitelist emails interceptor

### DIFF
--- a/lib/generators/rolemodel/mailers/mailers_generator.rb
+++ b/lib/generators/rolemodel/mailers/mailers_generator.rb
@@ -40,6 +40,42 @@ module Rolemodel
       end
     end
 
+    def add_production_mailer_defaults
+      prepend_to_file 'config/environments/production.rb', "require Rails.root.join('app/mailers/staging_mailer_interceptor')\n"
+
+      inject_into_file 'config/environments/production.rb', after: "config.action_mailer.perform_caching = false\n" do
+        optimize_indentation <<~'RUBY', 2
+          host = if ENV['REVIEW_APP'] == 'true' && ENV['HEROKU_APP_NAME'].present?
+            "https://#{ENV['HEROKU_APP_NAME']}.herokuapp.com"
+          else
+            ENV['PRODUCTION_HOST']
+          end
+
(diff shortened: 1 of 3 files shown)


lib/generators/rolemodel/mailers/mailers_generator.rb
```diff
@@ -40,6 +40,42 @@ def add_action_mailer_asset_host
...
+    def add_production_mailer_defaults
+      prepend_to_file 'config/environments/production.rb', "require Rails.root.join('app/mailers/staging_mailer_interceptor')\n"
+
+      inject_into_file 'config/environments/production.rb', after: "config.action_mailer.perform_caching = false\n" do
+        optimize_indentation <<~'RUBY', 2
+          host = if ENV['REVIEW_APP'] == 'true' && ENV['HEROKU_APP_NAME'].present?
+            "https://#{ENV['HEROKU_APP_NAME']}.herokuapp.com"
+          else
+            ENV['PRODUCTION_HOST']
+          end
```
calebwoods, line 52: This is coming from another generator right?

lib/generators/rolemodel/mailers/templates/app/mailers/staging_mailer_interceptor.rb
```diff
@@ -0,0 +1,8 @@
+class StagingMailerInterceptor
+  def self.delivering_email(message)
+    whitelisted_emails = ENV['WHITELISTED_EMAILS'].split(',').map(&:strip)
+    message.to = message.to & whitelisted_emails
+    message.subject = "STAGING - #{message.subject}"
```
calebwoods, line 5: Might want this also be able to be `REVIEW_APP`. I think that could be pulled from the ENV

mark-kraemer:
Add image fix width to 100px

PR #86, opened 2022-03-01
[TR #58] Buttons Revisited

Jeremy-Walton:
## Task

[TR #58](https://trello.com/c/DyO1hBwF)

## Why?

Structure the file using a `%btn` with base styles that extend `%btn` for button variants. Also, some of our button variants were outdated and needed some cleanup.

## What Changed

* [X] Change max radius token to use a large pixel value. This avoids strange behavior where % causes over rounding.
* [X] Change the default border width to 1px and remove the sm variant. Larger can be added on an as-needed by project basis.
* [X] Remove unnecessary base a tag style specificity.
* [X] Use `%btn` pattern for buttons to simplify usage (e.g. `.btn--primary` can be used without also needing `.btn`).
* [X] Prevent text color change on button hover
* [X] Make outlined button hover more subtle
* [X] Change `.btn--rounded` to `.btn--pill` since buttons are rounded (--radius 12px) by default and a "fully rounded" look is more like a pill.
* [X] Add large button variant
* [X] Adjust button font sizes and padding to scale better

## Screenshots

Old:
Screen Shot 2022-03-01 at 6 41 29 PM
Screen Shot 2022-03-01 at 6 41 35 PM
Screen Shot 2022-03-01 at 6 41 38 PM

New:
Screen Shot 2022-03-01 at 6 53 34 PM
Screen Shot 2022-03-01 at 6 53 38 PM

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables/semantic_color_scales.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables/semantic_color_scales.scss
@@ -10,7 +10,7 @@
     --color-#{$color}-lightest: var(--color-#{$color}-95);
     --color-#{$color}-lighter:  var(--color-#{$color}-80);
     --color-#{$color}-light:    var(--color-#{$color}-70);
-    --color-#{$color}-hover:     var(--color-#{$color}-50);
+    --color-#{$color}-hover:    var(--color-#{$color}-50);
     --color-#{$color}-base:     var(--color-#{$color}-40);
     --color-#{$color}-dark:     var(--color-#{$color}-30);
     --color-#{$color}-darker:   var(--color-#{$color}-20);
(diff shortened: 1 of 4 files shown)


lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables/semantic_color_scales.scss
```diff
@@ -10,7 +10,7 @@
     --color-#{$color}-lightest: var(--color-#{$color}-95);
     --color-#{$color}-lighter:  var(--color-#{$color}-80);
     --color-#{$color}-light:    var(--color-#{$color}-70);
-    --color-#{$color}-hover:     var(--color-#{$color}-50);
+    --color-#{$color}-hover:    var(--color-#{$color}-50);
```
scriswell, line 13: @Jeremy-Walton This one seems a bit limiting in practice. I wonder if we need a hover for each semantic color in the range? Something worth discussing anyway.

scriswell (approved):
@Jeremy-Walton Approved with just one comment.

lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/variables/semantic_color_scales.scss
Jeremy-Walton: Yeah, hover is extremely specific and not used outside of buttons.

Basically, `btn--primary` uses `--color-primary-base`, but for a hover effect, jumping to `--color-primary-light` is too drastic of a jump which is why this was added.

Instead of adding the special hover semantic name, I could just use the scale color in the hover css for buttons. That seems not great either but it does avoid adding a single-use semantic token

Current:
```
.btn--primary {
  @extend %btn;

  background: var(--color-primary-base);
  color: var(--color-on-primary-base);

  &:hover {
    background: var(--color-primary-hover);
  }
}
```

Proposed:
```
.btn--primary {
  @extend %btn;

  background: var(--color-primary-base);
  color: var(--color-on-primary-base);

  &:hover {
    background: var(--color-primary-50);
  }
}
```

PR #87, opened 2022-03-10
Fix visited link buttons and refactor to use loops

Jeremy-Walton:
## Why?

After the button updates recently merged, a tags using any of the button styles did not have the correct text color when the :visited css was being applied.

## What Changed

* [X] Fix visited link buttons text color
* [X] Refactor styles to use a scss loop

## Screenshots

Old:
Screen Shot 2022-03-10 at 2 06 57 PM

New:
Screen Shot 2022-03-10 at 2 06 45 PM

### DIFF
--- a/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/button.scss
+++ b/lib/generators/rolemodel/css/base/templates/app/javascript/stylesheets/components/button.scss
@@ -35,6 +35,10 @@
   text-decoration: none;
   color: var(--color-black);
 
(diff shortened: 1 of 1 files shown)


Jeremy-Walton:
I need to update the styleguide file to match the new styles.

Jeremy-Walton:
This is no longer relevant as the styles have been extracted into https://github.com/RoleModel/rolemodel-design-system

There is still a task left to make the generator use that package, but this particular PR is no longer needed.

PR #90, opened 2022-09-29
Apply Source Maps for Support

KXEinc:
## Task

[TR #XXX]()

## Why?

Why were the changes needed? What issues were the changes addressing?
(Note: some changes may seem unrelated to the ticket, this is a great place to explain further.)

We want to get better dev experience by ability to get source map natively without exposing it to non-RoleModel users in production environment. This Rack middleware serves source map only for allowed users.

## What Changed

What changed in this PR?

* [x] Add Rack middleware for serving js source maps
* [x] Configure WebPack to generate maps, append maps sources to application.js, and move maps to the different folder
* [x] Order generators by alphabet

## Screenshots


lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    private
+
+    def serve?(env)
+      ENV['RAILS_ENV'] == 'production' && env['PATH_INFO'].match?(/\.map\z/) && permit_user?(env)
+    end
+
+    def permit_user?(env)
+      (current_user_email = current_user(env)&.email) &&
+        (current_user_email.include?('@rolemodelsoftware.com') || @allowed_users_emails.include?(current_user_email))
```
tingraldi, line 26: I like the blanket permission for `@rolemodelsoftware.com`

lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+    def current_user(env)
```
tingraldi, line 28: Need a blank line between methods

lib/generators/rolemodel/source_map/source_map/config/initializers/source_map.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  #RoleModel::SourceMap(allowed_users_emails[, **options])
+  #allowed_users_emails should be a String or a Array of Strings
```
tingraldi, line 5: Grammatical error: Should be `an Array` instead of `a Array`

lib/generators/rolemodel/source_map/source_map_generator.rb
```diff
@@ -0,0 +1,11 @@
+#frozen_string_literal: true
+
+module Rolemodel
+  class ReadmeGenerator < Rails::Generators::Base
```
tingraldi, line 4: Shouldn't the class name be `SourceMapGenerator`?

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
```diff
@@ -2,4 +2,20 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 
 const environment = require('./environment')
 
+// This will cause source maps to be generated and a file:/// sourceMappingURL
```
tingraldi, line 5: The comment should be updated since the sourceMappingURL won't be a `file:///` URL.

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
```diff
@@ -2,4 +2,20 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
...
+// will be appended to the bundled JS
+environment.config.devtool = false
+const { SourceMapDevToolPlugin } = require('webpack')
+environment.plugins.prepend(
+  'SourceMapDevToolPlugin',
+  new SourceMapDevToolPlugin({
+    append: "\n//# sourceMappingURL=[url]",
```
tingraldi, line 12: I'm fairly certain that this will need to be

```javascript
   append: "\n//# sourceMappingURL=/maps/[url]",
```

since [url] will be substituted here with the name of the map file.

tingraldi (dismissed):
This will be very handy.

I've added some comments, but I did not give approval. I think someone that has built generators before should review this for approval.

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
KXEinc: It was intentional. So, if this comment contains only the file name, web-dev-tool will try to fetch the source map from the same directory as the original js bundle file. I don't want to expose `maps` folder to everyone. So, the middleware takes care of mapping URL to the `maps` folder.

lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
KXEinc: I added this 'wild card' because our support accounts have inconsistent email prefixes. This check eliminates the manual configuration step. From a security standpoint, this is anti-patter, but most of our apps have only one user with the "@rolemodelsoftware.com" domain (sure, this is a false statement). So, this should not be a problem, but I would like to discuss it.

lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
KXEinc: Resolved.

lib/generators/rolemodel/source_map/source_map/config/initializers/source_map.rb
KXEinc: Resolved.

lib/generators/rolemodel/source_map/source_map_generator.rb
KXEinc: It should.
Resolved.

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
tingraldi: Nice! I didn't catch that behavior in the middleware class.

lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
KXEinc: We decided not to go on this path. So, I will remove this condition.

tingraldi: On second thought, some of our deployed systems (like LightningCAD DPQ) allow users to register without validating email address ownership. So, blanket allowance for @rolemodelsoftware.com would be a problem.

KXEinc:
Please, review this PR and share your thought with me about this feature.

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
tingraldi: One more thing...

The sourceMappingURL form that is in place here is the same as it would be with

```javascript
environment.config.devtool = 'source-map'
```

So, there is no need to use the `SourceMapDevToolPlugin` to achieve this outcome.

Also, one downside of using the middleware approach to guard the source maps is that this mechanism won't work for situations where we don't have a logged in user. This would be the case for the LightningCAD DPQ guest editor and may apply to other applications as well. For those situations, the source maps would have to be downloaded and manually applied.

lib/generators/rolemodel/all_generator.rb
```diff
@@ -3,18 +3,19 @@ class AllGenerator < Rails::Generators::Base
     source_root File.expand_path('templates', __dir__)
 
     def run_all_the_generators
+      generate 'rolemodel:css:all'
```
Jeremy-Walton, line 6: What's the reasoning for changing the order of all these generators?

KXEinc: @Jeremy-Walton, I just like how alphabetic order looks. 
BTW, can you tell me if the running order of generators matters here?

Jeremy-Walton: It does. Some generators depend on other ones being run first. I.E. CSS depends on icons I think.

All that being said, RoleModel Rails is hugely out of date. The CSS does not reflect our latest learnings (doesn't pull in the design system), and it definitely does not support Rails 7

lib/generators/rolemodel/webpacker/templates/generated/config/webpack/development.js
KXEinc: @tingraldi, we need to obtain a user ID to confirm that the current user can see the source map. I believe the only IP address could be used as a user ID in the 'guest' case. But the problem is most internet providers assign dynamic IP addresses. So, developers must update the 'allowed IP' list whenever new IP addresses are assigned.  It is possible to build a script for updating the 'allowed IP' list, but I don't think it is worth it (We do not have a lot of cases like the 'LightningCAD DPQ guest editor' one). So, I agree with downloading and applying manually to web-dev-tool in similar cases.

KXEinc: Oh, I think that I should move WebPack configs to the production environment =)

tingraldi: Correct!

KXEinc: Resolved.

lib/generators/rolemodel/all_generator.rb
KXEinc: Ah, I see. Thank you for pointing me out!

KXEinc: Resolved.

lib/generators/rolemodel/source_map/source_map_generator.rb
Jeremy-Walton: Looks like this isn't resolved

lib/generators/rolemodel/source_map/source_map/app/middleware/rolemodel/source_map_generator.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Rolemodel
+  class SourceMapGenerator < Rails::Generators::Base
```
Jeremy-Walton, line 4: Is this supposed to be inherited from a rails generator? This seems like the actual middleware, not the generator.

Jeremy-Walton: Should this be in a templates folder?

lib/generators/rolemodel/source_map/source_map_generator.rb
```diff
@@ -0,0 +1,11 @@
...
+    source_root File.expand_path(__dir__)
+
+    def copy_source_map_files
+      copy_entry 'source_map', 'app'
```
Jeremy-Walton, line 8: Should this also be copying the actual source map middleware, not just the initializer?

Jeremy-Walton: Also, the js files don't seem to be getting copied in either.

Jeremy-Walton (changes requested):
Might want to try running this on a demo project to make sure it installs the files correctly. I'm not sure this is set up in a way that will actually work right now.

lib/generators/rolemodel/source_map/source_map_generator.rb
KXEinc: Resolved.

lib/generators/rolemodel/source_map/USAGE
```diff
@@ -0,0 +1,9 @@
+Description:
+    Adds Rack middleware for serving js source maps
+
+Example:
+    rails generate rolemodel:source_map
+
+    This will create:
+        app/middleware/rolemodel/source_map.rb
+        config/initializers/source_map.rb
```
Jeremy-Walton, line 9: THis can be removed

Jeremy-Walton (approved):
Approved with the removal of the initializer comment

PR #91, opened 2022-10-18
Rails 7

BenEinwechter:
## Task

[TR #84](https://trello.com/c/xLZycRmF/84-add-support-for-rails-7)

## Why?
Add and modify generators to setup a Rails v7 project.

## What Changed

- Remove obsolete Rails 5 code
- Replace Webpacker with Webpack v5 and replace babel with esbuild-loader for transpiling
- Remove `react_rails` and use Stimulus controller to inject React components into Rails views


mark-kraemer:
From Caleb:

Notes from when I'm running the generator on the RoleModel website being upgraded from Rails 6.
• Didn’t get the build commands generated in package.json
• I see the Honeybadger source maps was accounted for, but I’m not actually seeing the code for reporting errors. Might be worth talking through
• Didn’t seem to have an easy way to emit multiple CSS files so needed to modify the config and use of plugins
• Had to make some changes to the assets.rb initializer in order to be able to load SVGs

mark-kraemer:
@calebwoods I didn't have any issues compiling svgs in the test app and the assets.rb initializer remained unchanged
image

.ruby-version
```diff
@@ -1 +1 @@
-3.0.0
+3.1.3
```
calebwoods, line 1: Would bump to 3.1.4
```suggestion
3.1.4
```

Also not sure if would be benefit to going all the way to Ruby 3.2 at this point

README.md
```diff
@@ -84,7 +84,9 @@ bin/rails g
 
 ## Development
 
-We use the 2 embeded Rails apps [Rails 5 Example](./recreate_rails5_example) and [Rails 6 Example](./recreate_rails6_example) to test the usage of the generators.
+Install the versions of Node and Ruby specified in `.node-version` and `.ruby-version` on your machine. https://asdf-vm.com/ is a great tool for managing language versions. Then run `npm install -g yarn`.
+
+We use the embeded Rails apps [Rails 6 Example](./recreate_rails6_example) to test the usage of the generators.
 
 `recreate_rails6_example` is simply a fresh generated Rails 6 app. To recreate it use:
```
calebwoods, line 92: Should all this need to stay Rails 7?

example_rails7/.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.3
```
calebwoods, line 1:
```suggestion
3.1.4
```

calebwoods, line 1:
```suggestion
3.1.4
```

calebwoods: Although I guess this might be a generated file

lib/generators/rolemodel/README.md
```diff
@@ -17,7 +17,7 @@
 * [SoftDestroyable](./soft_destroyable)
 * [Source Map](./source_map)
 * [Testing](./testing)
-* [Webpacker](./webpacker)
```
calebwoods, line 20: Should this just be a rename not remove?

lib/generators/rolemodel/css/base/README.md
```diff
@@ -1,52 +1,10 @@
...
 ```
-rails g rolemodel:webpacker
 rails g rolemodel:css:all
 ```
 
-* View the styleguide modify as desired
-* Copy changes to generator with
+## What you get
+
+Pulls in the rolemodel optics design system
```
calebwoods, line 9: Should we just call this `rolemodel:optics` rather than CSS base?

calebwoods: I guess it is also doing slim, but that could be standalone

lib/generators/rolemodel/testing/rspec/templates/rails_helper.rb
```diff
@@ -60,4 +60,13 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  config.before :example, type: :system do
```
calebwoods, line 64: This something we need to do across the board? Might need to not which generators were changed as part of this effort in the readme so we know what we need to run for an upgrade

lib/generators/rolemodel/webpack/templates/app/assets/stylesheets/application.scss
```diff
@@ -0,0 +1,5 @@
+// This is a manifest file that'll be compiled into application.css, which will include
```
calebwoods, line 1: So this settles that for an upgrade we'll move all our stylesheets to `app/assets` and keep the javascript in `app/javascript` correct? Do we like that pattern?

lib/generators/rolemodel/webpack/webpack_generator.rb
```diff
@@ -0,0 +1,61 @@
...
+      File.write('package.json', JSON.pretty_generate(json) + "\n")
+    end
+
+    def add_npm_packages
+      say 'Move Webpack to devDependencies'
+      run 'yarn remove webpack webpack-cli'
+
+      say 'Add NPM packages for compiling JS and CSS'
+      js_packages = %w[
+        @honeybadger-io/webpack
```
calebwoods, line 18: Do we no longer have a standard `package.json` file? I'm wondering about where we bring in turbo, stimulus. Or should that be a separate generator

OutlawAndy: I think those are still coming from the turbo-rails & stimulus-rails gems and then imported in application.js, directly or indirectly.  Though I'm admittedly less familiar with the webpack approach used here.

.ruby-version
OutlawAndy: Ruby 3.2 has some nice additions.  Is there a reason not to upgrade to the current version?

lib/generators/rolemodel/testing/rspec/templates/rails_helper.rb
mark-kraemer: Leaving for now

lib/generators/rolemodel/webpack/webpack_generator.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def add_webpack_config
+      say 'Copying config files'
+
+      remove_file 'app/assets/stylesheets/application.css'
+      template 'app/assets/config/manifest.js', 'app/assets/config/manifest.js'
+      template 'app/assets/stylesheets/application.scss', 'app/assets/stylesheets/application.scss'
+
+      template 'app/javascript/application.js', 'app/javascript/application.js'
+      template 'app/javascript/controllers/index.js', 'app/javascript/controllers/index.js'
```
OutlawAndy, line 54: also should add the default 'app/javascript/controllers/application.js' file.  Alternatively, perhaps instead of copying any of these files we should instead run `rails turbo:install` & `rails stimulus:install`

Thoughts on this @mark-kraemer ?

calebwoods: If we aren't doing anything special with those generators we might just want to avoid having those in our generators. Simply running those installers should be enough

lib/generators/rolemodel/webpack/templates/app/javascript/application.js
```diff
@@ -0,0 +1,12 @@
+// Entry point for the build script in your package.json
+import '@hotwired/turbo-rails'
+import Honeybadger from '@honeybadger-io/js'
+import './controllers/index.js'
```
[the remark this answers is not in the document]
calebwoods: Should we just drop this?

lib/generators/rolemodel/optics/base/base_generator.rb
```diff
@@ -0,0 +1,35 @@
...
+    class BaseGenerator < Rails::Generators::Base
+      include Rolemodel::BundlerHelpers
+      source_root File.expand_path('templates', __dir__)
+
+      def add_optics_package
+        run 'yarn add @rolemodel/optics'
+      end
+
+      def add_slim
+        gem 'slim'
```
calebwoods, line 14: Does it still make sense for this to be mixed in with Optics? Could we separate it?

lib/generators/rolemodel/webpack/webpack_generator.rb
calebwoods: I'm wondering if we should do that or copy those files into our base webpack generator. Likely they are going to change, but also our React and other generators assume that stimulus is installed.

lib/generators/rolemodel/webpack/templates/app/javascript/application.js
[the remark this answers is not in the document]
calebwoods: If make Honeybadger an append to application.js that simplifies a lot of this code and makes upgrades easier

lib/generators/rolemodel/webpack/webpack_generator.rb
```diff
@@ -2,60 +2,88 @@ module Rolemodel
...
+      css-loader
+      css-minimizer-webpack-plugin
+      mini-css-extract-plugin
+      webpack-remove-empty-scripts
+    ]
+
+    def ensure_node_version
+      say "Ensuring Node version #{set_color(NODE_VERSION, :yellow)} is installed via nodenv"
+
+      run 'brew update && brew install nodenv node-build'
```
calebwoods, line 34: This is a bit heavy handed. I know most of us use nodenv, but would want to make this simple for someone using asdf not to get the setup broken.

That's the reason just sticking with the `.node-version` makes sense

lib/generators/rolemodel/slim/slim_generator.rb
```diff
@@ -0,0 +1,7 @@
+module Rolemodel
+  class SlimGenerator < Rails::Generators::Base
+    def add_slim
+      run 'bundle add slim'
```
OutlawAndy, line 4: The `gem` action does not check to see if it's already in your gemfile, while `bundle add` does.  `bundle add` also runs `bundle install` and accepts an optional `--group` argument.  So, I believe this is the better mechanism for adding gem dependencies.

calebwoods: 👍

lib/generators/rolemodel/optics/base/base_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Optics
+    class BaseGenerator < Rails::Generators::Base
+      source_root File.expand_path('templates', __dir__)
+
+      def add_optics_package
+        run 'yarn add @rolemodel/optics'
+      end
+
+      def add_slim
+        generate 'rolemodel:slim'
```
calebwoods, line 11: I kind of think we could remove this part from Optics entirely. Not anything specifically that depends on slim at this point as we don't ship a slim based styleguide like we used to

lib/generators/rolemodel/optics/base/templates/app/views/layouts/application.html.slim
```diff
@@ -0,0 +1,13 @@
+doctype html
```
calebwoods, line 1: This file should be in the slim generator not in Optics

lib/generators/rolemodel/optics/base/base_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      def remove_application_erb_file
```
calebwoods, line 14: This step and using the slim template should be in the slim generator

lib/generators/rolemodel/optics/base/README.md
```diff
@@ -0,0 +1,9 @@
+# Base CSS Generator
+
+```
+rails g rolemodel:optics:all
+```
+
+## What you get
+
+Pulls in the rolemodel optics design system and adds the slim gem
```
calebwoods, line 9:
```suggestion
Pulls in the RoleModel [Optics](https://docs.optics.rolemodel.design/) design system
```

lib/generators/rolemodel/optics/base/templates/app/views/layouts/application.html.slim
OutlawAndy: right, I forgot about that.  Thanks

lib/generators/rolemodel/optics/base/base_generator.rb
OutlawAndy: good call.  I added this reflexively, because it used to add the slim gem..

lib/generators/rolemodel/optics/base/templates/app/views/layouts/application.html.slim
Jeremy-Walton: That being said, it might be helpful to modify the application layout to work according to what Optics gives you. Technically not a requirement to use that though.

lib/generators/rolemodel/semaphore/templates/semaphore.yml.erb
```diff
@@ -3,7 +3,7 @@ name: <%= @project_name %>
 agent:
   machine:
     type: e1-standard-2
-    os_image: ubuntu1804
+    os_image: ubuntu2004
```
BenEinwechter, line 6: This file still is running a Webpacker compile command. 

Remove this whole block and any dependencies on it:
```yml
  - name: Assets Compile
    dependencies:
      - Build
    execution_time_limit:
      minutes: 20
    task:
      jobs:
        - name: Webpack Compile and Store Assets
          commands:
            - bundle exec rails webpacker:compile
            - cache store assets-public-$SEMAPHORE_WORKFLOW_ID public
```

README.md
```diff
@@ -84,18 +84,20 @@ bin/rails g
...
-bin/recreate_rails6_example
+bin/recreate_rails7_example
 ```
 
-Because Webpacker has so many moving parts, we want test bed where we can see the full [rolemodel:webpacker](./lib/generators/rolemodel/webpacker). To recreate it (after a version change, etc) use:
+Because Webpack has so many moving parts, we want a test bed where we can see the full [rolemodel:webpack](./lib/generators/rolemodel/webpack). To recreate it (after a version change, etc) use:
 
 ```shell
-bin/recreate_webpacker_example
+bin/copy_webpack_templates
```
BenEinwechter, line 100: This whole section is irrelevant since we're not using Webpacker anymore. Remove it and `bin/copy_webpack_templates` script.

README.md
```diff
@@ -84,18 +84,20 @@ bin/rails g
...
 
-`recreate_rails6_example` is simply a fresh generated Rails 6 app. To recreate it use:
+We use the embeded Rails apps [Rails 7 Example](./recreate_rails7_example) to test the usage of the generators.
+
+`recreate_rails7_example` is simply a freshly generated Rails 7 app. To recreate it use:
 
 ```shell
-bin/recreate_rails6_example
+bin/recreate_rails7_example
 ```
```
BenEinwechter, line 96: We should add a note here explaining how to test out a generator:

```suggestion
Then, cd into `example_rails7` and run a rolemodel generator to see how it affects a new Rails 7 project. For example:

cd example_rails7
bin/rails g rolemodel:webpack
```

example_rails7/app/javascript/application.js
```diff
@@ -0,0 +1,3 @@
+// Entry point for the build script in your package.json
+import "@hotwired/turbo-rails"
+import "./controllers"
```
BenEinwechter, line 3: This won't actually compile with our custom Webpack config because it requires file extensions. This file should be completely recreated by the webpack generator

```suggestion
import "./controllers/index.js"
```

example_rails7/app/javascript/controllers/index.js
```diff
@@ -0,0 +1,8 @@
+// This file is auto-generated by ./bin/rails stimulus:manifest:update
+// Run that command whenever you add a new controller or create them with
+// ./bin/rails generate stimulus controllerName
+
+import { application } from "./application"
```
BenEinwechter, line 5:
```suggestion
import { application } from "./application.js"
```

example_rails7/app/javascript/controllers/index.js
```diff
@@ -0,0 +1,8 @@
...
+
+import HelloController from "./hello_controller"
```
BenEinwechter, line 7:
```suggestion
import HelloController from "./hello_controller.js"
```

README.md
```diff
@@ -84,18 +84,20 @@ bin/rails g
...
+Install the versions of Node and Ruby specified in `.node-version` and `.ruby-version` on your machine. https://asdf-vm.com/ is a great tool for managing language versions. Then run `npm install -g yarn`.
 
-`recreate_rails6_example` is simply a fresh generated Rails 6 app. To recreate it use:
+We use the embeded Rails apps [Rails 7 Example](./recreate_rails7_example) to test the usage of the generators.
+
+`recreate_rails7_example` is simply a freshly generated Rails 7 app. To recreate it use:
 
 ```shell
-bin/recreate_rails6_example
+bin/recreate_rails7_example
```
BenEinwechter, line 94: The results of this command should be committed to git so we can see how the default Rails 7 files change from version to version. Right now when I run this generator, it changes a lot of the default files we already have in git which makes it hard to see what the generators are actually modifying.

BenEinwechter (commented):
Requesting changes.

example_rails7/app/javascript/controllers/index.js
OutlawAndy: I think, if these changes are required, then our generator may need to change, since this code is generated by Rails.

BenEinwechter: @OutlawAndy Yes, that's why I originally included those files in the webpack generator but I see you removed them in https://github.com/RoleModel/rolemodel_rails/pull/91/commits/718e998616f099779fbfc5925e839ba237c05f8e

lib/generators/rolemodel/optics/icons/templates/app/helpers/icon_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+      title: hover_text
+    }
+
+    options[:style] = "color: var(--op-color-#{color}-base);" if color # primary, neutral, alerts-danger, alerts-warning, alerts-notice, alerts-notice
+
+    tag.span(name, **options)
+  end
+
+  def icon(name, filled: false, size: 'medium', weight: 'normal', emphasis: 'normal', color: nil, classes: nil, hover_text: name)
+    using_custom_icon = custom_icon_path(name).present?
```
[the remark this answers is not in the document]
OutlawAndy: @Jeremy-Walton I appreciate the thoughtful feedback and agree with your suggestion, but I think I'd rather leave it for now and address this file directly in a separate PR.

 I didn't actually touch this file myself.  Just skimming it does strike me as significantly complex for what it's trying to accomplish..

example_rails7/app/javascript/controllers/index.js
OutlawAndy: @BenEinwechter - Ya, that came from a conversation w/ @calebwoods about reducing surface area and attempting to inject changes rather than replace files, so generators can be used on existing projects and not just new ones..

If every import statement that Rails generates will need to be modified in order to work with our JS strategy, we may need to rethink our JS strategy..  Maybe we could pair early next week to figure out how we move forward with this?

I don't want to merge something that doesn't work, but I'd rather merge an incremental improvement, than leave this branch open for several more weeks.

OutlawAndy: @BenEinwechter Is the issue just that webpack requires import statements to have file extensions?  If so, we could have a generator method that scans javascript files for import statements that don't have file extensions and assume that all they need is `.js` added to the path.  That would allow the generator to work with existing applications.  Even if there are edge cases where `.js` is not correct, the end user should be able to recognize these as they review generated changes..  Thoughts @calebwoods ?

Malachirwin:
During CCC Days I noticed that the readme still had the rails command at `rails new <app-name> --skip-test --database=postgresql` which technically breaks with rails 7 is installed since importmaps is the default JS bundling strategy. we might want to change this section on the readme to at least something that works. (The reason it fails is because there is no package json which means the webpack generator fails.)

OutlawAndy:
Caleb & Mark asked me to go ahead and merge this.  I'll be using it to upgrade Maverick and will open new PRs as problems arise.

PR #93, opened 2022-11-18
Fix Custom Icon Clip Paths

Malachirwin:
## Why?
In an app I realized that the clip path ids were being optimized to the same thing no matter if they were different in the file definition.

## What Changed
* [x] add the prefix id plugin which prefixes the ids with the base file name

## Screenshots (The issue is with the bottom icon's hatch pattern)
Before
Screen Shot 2022-11-18 at 3 53 13 PM
After
Screen Shot 2022-11-18 at 3 45 03 PM

### DIFF
--- a/lib/generators/rolemodel/css/icons/templates/config/webpack/loaders/svg-color-override-loader.js
+++ b/lib/generators/rolemodel/css/icons/templates/config/webpack/loaders/svg-color-override-loader.js
@@ -33,7 +33,17 @@ module.exports = function(content) {
         name: 'removeViewBox',
         active: false
       },
-      'removeDimensions'
+      'removeDimensions',
+      {
+        name: 'prefixIds',
+        params: {
+          delim: '-',
+          prefix: () => {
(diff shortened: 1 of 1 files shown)


jdmcleod (approved):

lib/generators/rolemodel/css/icons/templates/config/webpack/loaders/svg-color-override-loader.js
```diff
@@ -33,7 +33,17 @@ module.exports = function(content) {
...
       },
-      'removeDimensions'
+      'removeDimensions',
+      {
+        name: 'prefixIds',
+        params: {
+          delim: '-',
+          prefix: () => {
+            const pathParts = this.resourcePath.split('/')
+            return pathParts[pathParts.length - 1].replace('.svg', '')
```
kylesmile, line 43: Since this runs in Node, you could use [`path.basename()`](https://nodejs.org/dist/latest-v18.x/docs/api/path.html#pathbasenamepath-suffix).

PR #97, opened 2023-02-01
[TR#400] Add blazer generator

reedrolemodel:
## Task

[TR #400](https://trello.com/c/myNMLvoD)

## Why?

Blazer is quickly becoming a tool that we reach for on almost every project. What is missing from our use of it? What common customizations or fixes can we package?

A few things we could/should do

Switch from passing SQL back and forth in Blazer (except where absolutely necessary) to eliminate SQL injection.

Could also package Blazer for non-admin reporting as used on BGEA & Maverick

And one last thing is to create a rolemodel_rails generator
## What Changed

What changed in this PR?

* [x] Added generator for blazer
* [x] Improve security by preventing executing raw sql from POST data

## Screenshots


lib/generators/rolemodel/saas/blazer/blazer_generator.rb
```diff
@@ -0,0 +1,71 @@
+require_relative '../../../bundler_helpers'
+
+module Rolemodel
+  module Saas
```
calebwoods, line 4: Saas?

calebwoods: Ok. I see that's nesting approach. Could see it simply being rolemodel:blazer

lib/generators/rolemodel/saas/blazer/blazer_generator.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      def add_views
+        directory 'app/views/reports'
+      end
+
+      def add_styles
+        copy_file 'app/assets/stylesheets/blazer.css'
+        copy_file 'app/assets/stylesheets/selectize.css'
+        append_to_file 'app/assets/config/manifest.js', "\n//= link blazer.css"
```
calebwoods, line 58: This will be different base on the CSS approach being used right?

lib/generators/rolemodel/saas/blazer/blazer_generator.rb
timirwin: ok, will change

lib/generators/rolemodel/saas/blazer/blazer_generator.rb
timirwin: I defer to @reedrolemodel on that one. I forget why we needed that.

reedrolemodel: Yes, this assumes `sprockets` is being used which is also an assumption of `blazer`. [This issue](https://github.com/ankane/blazer/issues/379) is for adding support for `jsbundling` and `cssbundling` to `blazer`. Until more gems support newer approaches, it seems likely that `sprockets` will be needed.

lib/generators/rolemodel/testing/rspec/rspec_generator.rb
```diff
@@ -12,9 +12,11 @@ def install_rspec
         end
         run_bundle
 
-        gem_group :test do
-          gem 'capybara'
-          gem 'webdrivers'
+        if File.readlines('Gemfile').grep(/capybara|webdrivers/).none?
+          gem_group :test do
+            gem 'capybara' unless File.readlines('Gemfile').grep(/capybara/).any?
+            gem 'webdrivers' unless File.readlines('Gemfile').grep(/webdrivers/).any?
```
timirwin, line 18: Need to update this now.

PR #98, opened 2023-04-10
TurboFrame modal/panel support

OutlawAndy:
## Task

[TR #XXX]()

## Why?

Update the model generator to support our current turbo-frame strategy & remove our dependence on UJS.

## What Changed

* [x] remove custom modal/panel/confirm javascript
* [x] add @rolemodel/turbo-confirm package dependency
* [x] add @rolemodel/turbo-confirm initialization code
* [x] add required layouts for modal/panel
* [x] add required "turbo:frame-missing" event handler to support turbo-rails v1.4.0 and up.
* [x] add @rolemodel/optics integration for modal/panel styles


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    rolemodel_rails (0.2.0)
+    rolemodel_rails (0.3.0)
```
OutlawAndy, line 4: this seemed to happen automatically when I ran `bundle` locally after cloning the repo.

lib/generators/rolemodel/modals/README.md
```diff
@@ -2,31 +2,33 @@
...
 ## Example delete confirmation link
-`= link_to "Test Confirm", delete_path(model), class: 'btn btn--primary', data: { confirm: 'Are you sure?', confirm_details: "This will delete #{model.name}. This action cannot be undone.", confirm_button: 'Yes, delete it' }`
+`= link_to "Test Confirm", delete_path(model), class: 'btn btn--primary', data: { turbo_confirm: 'Are you sure?', confirm_details: "This will delete #{model.name}. This action cannot be undone." }`
 
 The data attributes needed to make this work
-- `confirm` (required) - Just like rails basic confirm functionality. This will be header text. E.g. "Are you sure?"
+- `turbo_confirm` (required) - Just like rails basic confirm functionality. This will be header text. E.g. "Are you sure?"
 - `confirm_details` (optional, default: nil) - Provides additional details/text to the action the user about to take. E.g. "This will remove everything"
-- `confirm_button` (optional, default: "Yes, I'm sure") - Allows the button text affirming the action to be changed
+- additional content slots need to be configured. See @rolemodel/turbo-confirm on Github for details.
```
OutlawAndy, line 18: perhaps @rolemodel/turbo-confirm should be updated to support `confirm_button` as a default content-slot?

OutlawAndy: this is done

lib/generators/rolemodel/modals/README.md
```diff
@@ -2,31 +2,33 @@
 
 ## What you get
 
-* Custom confirms
-* Modal and panel styling
-* Modal and panel view and link helper
-* Modal and panel javascript
+* Custom confirm dialogs via `@rolemodel/turbo-confirm` integration
+* Modal and panel styling ..via `@rolemodel/optics` integration?
+* Modal and panel layouts w/ link helpers (or you could just, e.g. `link_to ... data: {turbo_frame: 'modal'}`)
```
OutlawAndy, line 7: It's my opinion that the link helpers should be removed entirely.

* I believe they should be removed because they obscure how modals and panels work, making them seem more magical than they really are.
* Links often require application/context specific customization.  Which is likely to lead to bloated link helpers as they are adapted to accommodate application specific use-cases.
* I think it's better to use `link_to` directly and ask developers to add `data-turbo-frame='modal'` attributes explicitly.

Does anyone else have an opinion on this?

lib/generators/rolemodel/modals/templates/app/helpers/dynamic_link_helper.rb
```diff
@@ -1,9 +1,9 @@
 module DynamicLinkHelper
   def panel_link_to(name, url, options)
-    link_to(name, url, options.merge(data: { remote: true, panel: true }))
+    link_to(name, url, options.merge(data: { turbo_frame: 'panel' }))
```
OutlawAndy, line 3: It's my opinion that the link helpers should be removed entirely.

* I believe they should be removed because they obscure how modals and panels work, making them seem more magical than they really are.
* Links often require application/context specific customization. Which is likely to lead to bloated link helpers as they are adapted to accommodate application specific use-cases.
* I think it's better to use link_to directly and ask developers to add data-turbo-frame='modal' attributes explicitly.

Does anyone else have an opinion on this?

lib/generators/rolemodel/modals/modals_generator.rb
```diff
@@ -2,34 +2,40 @@ module Rolemodel
...
+    end
+
     def install_icons
       generate 'rolemodel:css:base' unless File.exists?(Rails.root.join('app/javascript/packs/stylesheets.scss'))
       generate 'rolemodel:css:icons'
     end
 
     def add_ui_components
-      copy_file 'app/helpers/dynamic_link_helper.rb'
+      copy_file 'app/helpers/dynamic_link_helper.rb' # TODO: is this still useful?
```
calebwoods, line 20: This is the helper for adding the data attributes?

lib/generators/rolemodel/modals/templates/app/helpers/dynamic_link_helper.rb
calebwoods: We've got more complete versions that we've built into other projects that handle taking a block, etc

OutlawAndy: Feels to me like unnecessary complexity, just to add a single data attribute.  I'm not arguing against helpers, just that they're more project specific.  On NHCS for example, we added this one:

```ruby
  def modal_edit_button(
    resource,
    url: [:edit, resource],
    class: 'btn--small', data: {}
  )
    return unless policy(resource).update?

    icon_button 'pencil', url, class:, data: data.reverse_merge(turbo_frame: 'modal')
  end
```

calebwoods: What if we call them `ModelLinkHelper` and `PanelLinkHelper` and support taking a block. Gives us a pattern to build off

lib/generators/rolemodel/modals/templates/app/views/layouts/application.html.slim
```diff
@@ -7,13 +7,10 @@ html
     = stylesheet_pack_tag 'stylesheets', 'data-turbolinks-track': 'reload'
     = javascript_pack_tag 'application', 'data-turbolinks-track': 'reload'
   body.flex
+    = turbo_frame_tag 'modal'
```
calebwoods, line 10: Rather than doing a full file replace. Should just find the body and insert these lines

PR #101, opened 2023-06-04
Add Kaminari

Jeremy-Walton:
## Why?

[Kaminari](https://github.com/kaminari/kaminari) is a tool we commonly use for paginating data. Optics is adding some styling around pagination so this generator configures your application to use the correct styling.

## What Changed

* [X] Fix Optics All Generator to use the correct module
* [X] Add Kaminari generator

## Note

This relies on https://github.com/RoleModel/optics/pull/175 which is not released yet. It should be fine for the time being though


Jeremy-Walton:
Should this get put under the Optics section? It is dependent on styling optics provides, but not for the paging functionality

Issue #100, opened 2023-05-31
New rails 7 generator issues recently found


mark-kraemer:
#### Webpack Generator:
- [x] `The file /Users/markkraemer/Sites/rolemodel/engage-learn/app/javascript/application.js does not appear to exist` failure will break the generator and prevent it from continuing
- [x] `"build": "webpack --config webpack.config.js"` is not injected into the existing package.json file 
- [ ] `jsbundling-rails` gem needs to be added to the gemfile, followed by running `bin/rails javascript:install:webpack`
- [ ] `stimulus-rails` gem should be added to the gemfile.  This gem provides the stimulus generator & the `stimulus:manifest:update`  task for auto generating the app/javascript/controllers/index.js file
- [ ] app/javascript/controllers/application.js template (see below)

OutlawAndy:
app/javascript/controllers/application.js

```JS
import { Application } from "@hotwired/stimulus"

const application = Application.start()

application.debug = process.env.NODE_ENV === 'development'
window.Stimulus   = application

export { application }

```

Braden-077:
- [ ] Fix Mailer Generator (issues with styling)
- [ ] Update Semaphore Generator with ubuntu2004

BakoviDagi:
I could just be doing things wrong, but when I run the `rolemodel:all` on a fresh project, I get an error

```
An error occurred while installing bcrypt (3.1.19), and Bundler cannot continue.

In Gemfile:
  devise was resolved to 4.9.2, which depends on
    bcrypt
    generate  devise:install
       rails  generate devise:install
Could not find bcrypt-3.1.19 in locally installed gems
Run `bundle install` to install missing gems.
```

Jeremy-Walton:
#107 Is now merged which addressed some of these things.

@BakoviDagi Yeah things unfortunately don't just "work" out of the box. Not sure what causes it but definitely worth investigation.

OutlawAndy:
I'm going to close this since it's almost 2 years old.  We should track individual issues as they pop up.  Feel free to `@` me if you don't get a response from anyone else.

PR #104, opened 2023-06-23
Modals Generator

Jeremy-Walton:
## Task

Fix modal issues noted in #100 

## Why?

The modals generator was not pointing to the correct files and the template was not up to date.
Since Optics provides the modal as a full component now instead of an addon, these needed to be updated.

There might be a future PR to start using the `dialog` element, but there is further discussion needed on it.

## What Changed

* [X] Clean up injected styles
* [X] Use Optics template for modal

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    rolemodel_rails (0.6.0)
+    rolemodel_rails (0.6.1)
 
 GEM
   remote: https://rubygems.org/
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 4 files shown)


OutlawAndy (approved):
Thanks @Jeremy-Walton! 👍

Jeremy-Walton:
@OutlawAndy I realized there is more work to be done getting the Confirm Dialog updated to use the latest from Optics as well. https://docs.optics.rolemodel.design/?path=/docs/components-confirm-dialog--docs

Would love to pair with you on that as well. Maybe next week some time.

Braden-077:
@Jeremy-Walton What's the status on this PR?

Jeremy-Walton:
@Braden-077 Would like to see some feedback from @calebwoods on https://trello.com/c/llL8ASEM/85-migrate-to-the-dialog-element first but hadn't gotten a chance to bring it up with him.

PR #105, opened 2023-07-02
Add switch checkbox control

Jeremy-Walton:
## Why?

Optics has a new component releasing in the next version. In preparation for that, we want to add a simple form custom input and wrapper to utilize it better.

## What Changed

* [X] Add custom input
* [X] Add custom wrappers
* [X] Add usage documentation

## Screenshots

Screenshot 2023-07-02 at 6 24 55 PM


Jeremy-Walton:
This is actually waiting on some potential changes in the unreleased Optics Implementation. https://github.com/RoleModel/optics/pull/185

lib/generators/rolemodel/simple_form/templates/app/inputs/switch_checkbox_input.rb
```diff
@@ -0,0 +1,34 @@
...
+# <%= f.input :my_field, as: :switch_checkbox %>
+# <%= f.input :my_field, as: :switch_checkbox, small: true %>
+# <%= f.input :my_field, as: :switch_checkbox, label_after_input: true %>
+# <%= f.input :my_field, as: :switch_checkbox, wrapper: :switch_wrapper %>
+class SwitchCheckboxInput < SimpleForm::Inputs::BooleanInput
+  def input(wrapper_options = nil)
+    merged_input_options = merge_wrapper_options(input_html_options, wrapper_options)
+
+    switch_group = template.content_tag(:div, class: "switch #{'switch--small' if options[:small]}") do
+      build_check_box_without_hidden_field(merged_input_options) +
```
OutlawAndy, line 20: Is this related to the `include_hidden: false` flag that the `check_box` form helper exposes.

Usually that hidden field is important, so just checking that the switch produces the same underlying form inputs that the form builder does.

Jeremy-Walton: The html this generates looks like the following:

```html
<div class='form-group form-group--inline'>
  <div class="switch">
    <input type="checkbox" id="switch-demo" >
    <label for="switch-demo"></label>
  </div>
  <label class="form-label" for="switch-demo">Switch Text</label>
</div>
```

Since it already has a checkbox in there, it won't need the form builder hidden field

Jeremy-Walton: I tested in a demo application and everything worked great!

OutlawAndy: Just want to make sure we're speaking the same language.  Checkboxes which are "unchecked" do not submit their value to the server.  Since we often use them in Rails apps to represent a Boolean column, we also need a hidden input with the same name and the inverse value to come before the rendered checkbox in the form.  The hidden input is overwritten by the checkbox when it is is checked, because it has the same name.  When the checkbox is not checked, only the hidden input's value is submitted.  Without the hidden input, a checkbox can set a Boolean value to "true", but it cannot later update it back to "false".

Is this component meant for a different use-case than the one mentioned above?

Jeremy-Walton: Ah yeah I didn't fully test it. You are correct.

I've updated the input helper. The outputted HTML would look like 

```
<div class="switch ">
  <input name="project[test]" type="hidden" value="0" autocomplete="off">
  <input type="checkbox" value="1" checked="checked" name="project[test]" id="project_test">
  <label class="switch_checkbox optional form-label" for="project_test">Test</label>
</div>
```

And it looks like 
Screenshot 2023-08-11 at 11 14 00 AM

(though this is dependent on the options passed)

OutlawAndy (approved):

PR #111, opened 2023-10-20
Fix a lot of broken or incorrect generators

### LINKED ISSUE #106: Steps Needed to make the Generator Work in a new project
1. Verify you're using rails v7
2. `rails new <app> -j=webpack -c=sass -d=postgresql -T`
3. `cd <app>`
4. `bundle add bcrypt`
5. Add generator to Gemfile
6. `bundle`
7. Run the generator(s)
8. `bin/setup`
9. Start the server

Jeremy-Walton:
## Why?

Following the instructions in the README and running the generate all command was not working out of the box. This aims at fixing a majority of the issues.

Fixes #106 
Addresses some things in #100

## What Changed

* [X] Make the React Generator Optional. Since a lot of our newer projects don't use React, running the all command will ask if you want it.
* [X] Update the eslint generator to use the latest ecmascript setting. Fixed the injected package command. Also cleaned up some of our templates to not generate with warnings.
* [X] Fix semaphore generator yml commands for webpack
* [X] Update the README to instruct with the correct  `rails new` command and also list all the generators that exist.
* [X] Fix the Devise saas generator to not crash when installing bcrypt.
* [X] Fix the rubocop generator to put the custom cop in `lib/cops/...` instead of app since the linter expected it to be in lib.
* [X] Add rubymines idea folder to `.gitignore` file


Braden-077 (approved):
My only comment on this is that we should add one directory to the `.gitignore` in the Github generator. `/.idea` is a RubyMine specific set of files which is obnoxious and irrelevant.

BakoviDagi (approved):

mark-kraemer (approved):

PR #117, opened 2024-04-18
Allow for nullify to cascade on soft destroy

Braden-077:
## Why?

This PR adds the cascade effect to dependencies with `dependent: :nullify` by nullifying the dependency on soft destroy.

## What Changed

* [x] Add nullify to soft destroy (following what Rails nullify does)

### DIFF
--- a/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
+++ b/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
@@ -8,6 +8,7 @@ module SoftDestroyable
     scope :only_deleted, -> { where.not(deleted_at: nil) }
     define_model_callbacks :soft_destroy
     define_model_callbacks :restore
+    after_soft_destroy :cascade_soft_destroy_nullify
   end
 
   module ClassMethods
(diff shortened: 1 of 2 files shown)


lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
```diff
@@ -70,9 +70,20 @@ def restore!
     restore(deleted_at) || raise_validation_errors
   end
 
+  def cascade_soft_destroy_nullify
```
justwiebe, line 73: Hmm, this would come in handy for a specific case in Gimme Credit https://github.com/RoleModel/gimme-credit/pull/495/files

Is there a reason you aren't checking `belongs_to` and `has_one` associations?

@BlaineIrvin perhaps we should use this pattern in Gimme Credit?

Braden-077: @justwiebe No, the `has_one` would be another good one to include! 

I'm curious why we'd need to do anything to our `belongs_to`, could you explain that one?

justwiebe: `belongs_to` might be an even bigger edge case, but for our situation, we've got a chain of reports that "replace" the previous one. We disallow soft destroying a report in the middle of the chain to side step complications, but we do allow soft destroying the latest report. To keep the chain linear, we decided to remove the soft destroyed one from it altogether be removing the `belongs_to` association

Braden-077: It sounds like you're having the child class nullify its own association. There could CERTAINLY be value in that one!

Braden-077: That sounds like an `after_soft_destroy nullify_self`

Braden-077: Well sure, maybe not out of the box. You could use an arbitrary value on that dependent call. 😉

Braden-077: @justwiebe @BlaineIrvin Thinking back on this - I'm going to add the `has_one` update. I think the specific case of the parent record might be best solved by touching the parent record.

PR #118, opened 2024-04-22
Add LightningCAD Generator

Malachirwin:
## Why?
The former system for spinning up a L.CAD app doesn't work in the slightest and was not worth repairing. This pulls some pieces from that but is implemented on `rolemodel_rails`

## What Changed
* [x] Pull required setup from the old generator
* [x] Split the generators into install, test, and webpack
* [x] Add ChromeCAD support out of the box

### What the generator adds
- Global setup
  - `.npmrc`
  - `.eslintrc.js`
- L.CAD Setup
  - Initializers
    - `javascript/config/initializers/smartJSON.js`
  - React components
    - `javascript/components/LocalIconFactory.jsx`
    - `javascript/components/MaterialIcon.jsx`
- Demo editor
  - `controllers/editor_controller.rb`
  - `views/editor/editor.html.slim`
  - `javascript/components/App.jsx`
- Jasmine test setup
  - `jasmine.json`
  - `spec/javascript/shared/TestSetup.js`
  - `spec/javascript/shared/testSpec.js`

## Later
This doesn't fully support the setup for persisted project setup which is most of the use case we need to setup. This only replaces the old generators that don't work. After we have the replacement we can look at some of those other bells and whistles.


lib/generators/lightning_cad/templates/.eslintrc.js
```diff
@@ -0,0 +1,123 @@
+export default {
```
kylesmile, line 1: Probably better to just use the ESLint config provided by the generic (non-L.CAD) generator. There might be changes from projects that could be good to bring in to our base config.

lib/generators/lightning_cad/test_generator.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module LightningCad
+  class TestGenerator < Rails::Generators::Base
+    source_root File.expand_path('./templates', __dir__)
+
+    def add_jasmine_tests
+      say 'Adding jasmine'
+      run "yarn add --dev jasmine@^5.1.0"
+
+      run 'npm pkg set scripts.test_shared="NODE_ENV=test NODE_PATH="./node_modules:./app/javascript:$NODE_PATH" jasmine --config=jasmine.json"'
```
kylesmile, line 11: I don't think NODE_PATH works with ESM, but we can add a custom import hook to replicate the behavior.

lib/generators/lightning_cad/webpack_generator.rb
```diff
@@ -0,0 +1,100 @@
...
+      JS
+
+      insert_into_file 'webpack.config.js', loaders, after: "rules: [\n"
+    end
+
+    def update_esbuild_loader
+      say 'Updating esbuild loader in the config'
+
+      insert_into_file 'webpack.config.js', "        include: /app\\/javascript|@rolemodel\\/lightning-cad/,\n", before: "        loader: 'esbuild-loader',"
+      gsub_file 'webpack.config.js', 'es2021', 'esnext'
```
kylesmile, line 77: We probably don't want to change the output target.

lib/generators/lightning_cad/webpack_generator.rb
```diff
@@ -0,0 +1,100 @@
...
+      },
+      JS
+
+      insert_into_file 'webpack.config.js', loaders, after: "rules: [\n"
+    end
+
+    def update_esbuild_loader
+      say 'Updating esbuild loader in the config'
+
+      insert_into_file 'webpack.config.js', "        include: /app\\/javascript|@rolemodel\\/lightning-cad/,\n", before: "        loader: 'esbuild-loader',"
```
kylesmile, line 76: I don't think this is needed, since the base config doesn't seem to exclude node_modules from being processed with esbuild-loader.

kylesmile (commented):
One thing we'll want to keep in mind is that this repo is public at the moment, whereas LightningCAD is private. We'll want to be mindful of how much L.CAD-specific code ends up in this repo.

OutlawAndy:
@Malachirwin Looks like there is a lot here.  Since LightningCAD apps are a thing unto themselves, would it make sense to convert this into a stand alone [application template](https://guides.rubyonrails.org/rails_application_templates.html) instead? cc @kylesmile

PR #120, opened 2024-05-02
Add GoodJob generator

timirwin:
May 2024 CCC days

GoodJob Background Jobs best practices turned into a generator for v3 defaults

### DIFF
--- a/lib/generators/rolemodel/good_job/good_job_generator.rb
+++ b/lib/generators/rolemodel/good_job/good_job_generator.rb
@@ -1,0 +1,74 @@
+require_relative '../../bundler_helpers'
+
+module Rolemodel
+  class GoodJobGenerator < Rails::Generators::Base
+    include Rolemodel::BundlerHelpers
+    source_root File.expand_path('templates', __dir__)
+
+    def install_good_job
+      say 'Install GoodJob'
+
+      bundle_command 'add good_job'
+      run_bundle
+
+      generate 'good_job:install'
+    end
+
+    def configure_good_job
+      say 'Add GoodJob configuration'
+
+      inject_into_file 'config/application.rb', after: "class Application < Rails::Application\n" do
+        optimize_indentation <<~'RUBY', 4
+          # Add GoodJob ActiveJob queue adapter
+          config.active_job.queue_adapter = :good_job
+
+        RUBY
+      end
+
+      gsub_file 'config/environments/production.rb', 'config.active_job.queue_adapter', '# config.active_job.queue_adapter'
+    end
+
+    def configure_good_job_routes
+      say 'Add GoodJob routes'
+
+      route_text = <<~RUBY
+        # Example but update per the authorization strategy of app
+        namespace :admin do
+          # authenticate :user, ->(user) { user.admin? } do
+            mount GoodJob::Engine => 'good_job'
+          # end
+        end
+      RUBY
+      route route_text
+    end
+
+    def configure_procfile
+      say 'Add GoodJob to Procfile'
+
+      if File.exist?(Rails.root.join('Procfile'))
+        append_to_file 'Procfile', <<~COMMAND
+          worker: bundle exec good_job start
+        COMMAND
+      end
+
+      append_to_file 'Procfile.dev', <<~COMMAND
+        worker: bundle exec good_job start
+      COMMAND
+    end
+
+    def copy_initializers
+      say 'Add GoodJob initializers'
+
+      copy_file 'config/initializers/active_job.rb'
+      copy_file 'config/initializers/good_job.rb'
+    end
+
+    def finishing_notes
+      say <<~NOTES
+        *** Reminder to update Honeybadger gem to version 5.7.0 or later to get correct GoodJob error notifications in Honeybadger
+
+        *** Reminder to also update your job classes to include appropriate concurrency controls (enqueue_limit/perform_limit with keys)
+      NOTES
+    end
+  end
+end
--- a/lib/generators/rolemodel/good_job/templates/config/initializers/active_job.rb
+++ b/lib/generators/rolemodel/good_job/templates/config/initializers/active_job.rb
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
+# *** NOTE ABOUT ORDERING OF discard_on/retry_on
+# https://api.rubyonrails.org/v7.1.3.2/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
+# 'retry_on' and 'discard_on' handlers are searched from bottom to top, and up the class hierarchy.
+# The handler of the first class for which exception.is_a?(klass) holds true is the one invoked, if any.
+
 # Add retry at ActiveJob::Base level to ensure all jobs retry 25 times by default regardless
 # of where the job originates from
 # Note: Rails 6 should use :exponentially_longer
 ActiveJob::Base.retry_on StandardError, wait: :polynomially_longer, attempts: 25
+
+# Automatically retry jobs that encountered a deadlock
+ActiveJob::Base.retry_on ActiveRecord::Deadlocked
 
 ActiveJob::Base.discard_on ActiveJob::DeserializationError
 ActiveJob::Base.discard_on ActiveRecord::RecordNotFound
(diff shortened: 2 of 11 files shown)


lib/generators/rolemodel/good_job/templates/config/initializers/active_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Add retry at ActiveJob::Base level to ensure all jobs retry 25 times by default regardless
+# of where the job originates from
+# Note: Rails 6 should use :exponentially_longer
+ActiveJob::Base.retry_on StandardError, wait: :polynomially_longer, attempts: 25
+
```
timirwin, line 7:
```suggestion

# Automatically retry jobs that encountered a deadlock
ActiveJob::Base.retry_on ActiveRecord::Deadlocked

```

lib/generators/rolemodel/good_job/templates/config/initializers/active_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
```
timirwin, line 2:
```suggestion

# *** NOTE ABOUT ORDERING OF discard_on/retry_on
# https://api.rubyonrails.org/v7.1.3.2/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
# 'retry_on' and 'discard_on' handlers are searched from bottom to top, and up the class hierarchy.
# The handler of the first class for which exception.is_a?(klass) holds true is the one invoked, if any.

```

lib/generators/rolemodel/good_job/good_job_generator.rb
```diff
@@ -0,0 +1,74 @@
...
+    def configure_good_job
+      say 'Add GoodJob configuration'
+
+      inject_into_file 'config/application.rb', after: "class Application < Rails::Application\n" do
+        optimize_indentation <<~'RUBY', 4
+          # Add GoodJob ActiveJob queue adapter
+          config.active_job.queue_adapter = :good_job
+
+        RUBY
+      end
```
kylesmile, line 26: Is the preference to modify `application.rb` or include these kinds of settings in a file in `config/initializers.rb`? I've typically done the latter in most cases. I think we've found that can be more reliable for generators, too.

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  config.good_job = {
+    smaller_number_is_higher_priority: true, # new default in V4 and for all of ActiveJob
+    queues: ENV.fetch('GOOD_JOB_QUEUES', 'pdf:1;-pdf'),
+    execution_mode: Rails.env.test? ? :inline : :external
```
reedrolemodel, line 7: Aren't these defaults? Did you consider leaving them out?

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  config.good_job = {
+    smaller_number_is_higher_priority: true, # new default in V4 and for all of ActiveJob
+    queues: ENV.fetch('GOOD_JOB_QUEUES', 'pdf:1;-pdf'),
```
reedrolemodel, line 6: Did you consider adding a comment about the default value? I'm not sure what it does.

kylesmile, line 6: Should the generator assume an app will have a `pdf` queue?

kylesmile (approved):
A couple questions, but overall looks fine to me.

lib/generators/rolemodel/good_job/good_job_generator.rb
reedrolemodel: This approach [follows the README](https://github.com/bensheldon/good_job?tab=readme-ov-file#set-up).

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
kylesmile: I think the default for development mode is `:async`.

reedrolemodel: Ah, you're right. Which then raises the question: should the external process be added to the Procfile? Also curious why not use `:async` in development.

kylesmile: > should the external process be added to the Procfile.dev?

I think this generator _is_ doing that already.

> Also curious why not use `:async` in development.

I would assume to make development more like production.

reedrolemodel: Ah, I see that now

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
mikehale: I agree, this is rather cryptic and should be documented.

mikehale: IIRC this was intended mostly as an example that was extracted from several projects which do have a pdf queue. @jdmcleod @timirwin and I had some discussion about whether it should be called something else like `large_job`, or `serial_job`.

lib/generators/rolemodel/good_job/good_job_generator.rb
mikehale: @kylesmile that is a good observation. In our context an initializer probably makes more sense.

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
timirwin: We did consider leaving them out. We've had a couple of projects run into issues with `:async` in development. I think it is related to hot reloading when code changes are made causing things to hang. Instead of tracking that down, those projects have switched to `:external` for development. I've not experienced it myself but wanted to save people the troubles.

lib/generators/rolemodel/good_job/templates/config/initializers/good_job.rb
timirwin: Yes, I can add a comment.

timirwin: Our guiding principle was to minimize overrides. `GOOD_JOB_QUEUES` will be used by default (by GoodJob) and typically is set in Heroku config. Rather than have instructions on overriding it, we opted to give a slightly better default, in particular communicating some basic things about queues which are not used/understood AND handle our most common exception case, `pdf` generation using background jobs. I'll add a comment.

lib/generators/rolemodel/good_job/good_job_generator.rb
timirwin: @kylesmile Yeah, we were trying to follow the documentation so people would know where to look if they read the GoodJob docs. I'm open to moving it but in checking our RoleModel projects, only 2 (landone-takeoff and thundercloud) put it in an initializer. Everything else (141 instances) are either in config/application.rb (11) or config/environments/<something> (130) files because that's where Rails puts it by default.

kylesmile (approved):

lib/generators/rolemodel/good_job/good_job_generator.rb
jdmcleod: For what its worth, their docs indicate that you can keep it in the initializer if you change it from `config.active_job.queue_adapter = :good_job` to `ActiveJob::Base.queue_adapter = :good_job`. I didn't try that out though to test if it solved my bug since the other way worked.

tingraldi: > @jdmcleod [found out](https://rolemodelsoftware.slack.com/archives/C051DL755/p1729690474673799?thread_ts=1729606215.178869&cid=C051DL755) that putting `config.active_job.queue_adapter = :good_job` into an initializer, it caused issues with the jobs running inline instead of in the Heroku worker process when cron was added. We should consider moving that in [thundercloud](https://github.com/rolemodel/thundercloud) @tingraldi @seanpk too.

This isn't biting us right now in thundercloud, but it would be good to relocate the config to avoid potential issues in  the future. I've added an [issue](https://linear.app/rolemodelsoftware/issue/DPQ-343/relocate-good-job-setting) to Linear for this.

PR #121, opened 2024-07-08
Fix Rspec rails helper asset precompile step

Jeremy-Walton:
## Why?

There is a very confusing side affect when you run `rails new` with `--skip-test` (since we use Rspec instead). In the latest versions of rails it will cause system tests to throw the following error

```shell
Unrecognized command "test:prepare" (Rails::Command::UnrecognizedCommandError)
Did you mean?  test:helpers
```

Using `--skip-test` generates `config/application.rb` with

```ruby
require_relative "boot"

require "rails"
# Pick the frameworks you want:
require "active_model/railtie"
require "active_job/railtie"
require "active_record/railtie"
require "active_storage/engine"
require "action_controller/railtie"
require "action_mailer/railtie"
require "action_mailbox/engine"
require "action_text/engine"
require "action_view/railtie"
require "action_cable/engine"
# require "rails/test_unit/railtie"
```

Notice that last commented out line...
If you don't skip tests, it generates

```ruby
require_relative "boot"

require "rails/all"
```

The issue is that test:prepare is a placeholder rake task brought in by `rails/test_unit/railtie` that rspec hooks into and adds the asset precompiling behavior to system tests. By skipping tests, it doesn't include this rake task which causes Rspec to fail.

## What Changed

* [X] Add directions to the Readme to add back the needed railtie require

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,16 +17,6 @@ The rolemodel_rails gem expects to be added to an existing Rails project. Typica
 rails new <app-name> --javascript=webpack --css=sass --database=postgresql --skip-test
 ```
 
-The Rspec generators require a railtie rake task to exist before running.
-
-Rspec hooks into `rails test:prepare` which doesn't exist due to using `--skip-test` in the new app generator. To add it back, remove all the imports at the top of `config/application.rb` and replace with:
-
-```ruby
-require_relative "boot"
-
-require "rails/all"
-```
-
 The Devise generator requires your database to exist before running.
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ The rolemodel_rails gem expects to be added to an existing Rails project. Typica
 rails new <app-name> --javascript=webpack --css=sass --database=postgresql --skip-test
 ```
 
+The Rspec generators require a railtie rake task to exist before running.
+
+Rspec hooks into `rails test:prepare` which doesn't exist due to using `--skip-test` in the new app generator. To add it back, remove all the imports at the top of `config/application.rb` and replace with:
+
+```ruby
+require_relative "boot"
+
+require "rails/all"
+```
+
 The Devise generator requires your database to exist before running.
 
 ```shell
--- a/lib/generators/rolemodel/testing/rspec/templates/rails_helper.rb
+++ b/lib/generators/rolemodel/testing/rspec/templates/rails_helper.rb
@@ -63,7 +63,7 @@ RSpec.configure do |config|
 
   config.before :example, type: :system do
     unless ENV['ASSET_PRECOMPILE_SUCCESSFUL']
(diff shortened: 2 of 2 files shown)


reedrolemodel:
> The issue is that test:prepare is a placeholder rake task brought in by rails/test_unit/railtie that rspec hooks into and adds the asset precompiling behavior to system tests. By skipping tests, it doesn't include this rake task which causes Rspec to fail.

Are you sure? It seems the other way around. That is, gems [like jsbundling-rails](https://github.com/rails/jsbundling-rails/blob/7faf59f85303d44f39e78cfce21db09dbe48d60a/lib/tasks/jsbundling/build.rake#L69) hook into `test:prepare`. `rspec-rails` [defines `spec:prepare`](https://github.com/rspec/rspec-rails/blob/8c17b4e5020a4d264e8a79e294c58b5c1ef2b005/lib/rspec/rails/tasks/rspec.rake#L11) (which `jsbundling-rails` also hooks into). Shouldn't the solution be to run `spec:prepare` instead of `test:prepare`?

Braden-077 (approved):

Braden-077:
> > The issue is that test:prepare is a placeholder rake task brought in by rails/test_unit/railtie that rspec hooks into and adds the asset precompiling behavior to system tests. By skipping tests, it doesn't include this rake task which causes Rspec to fail.
> 
> Are you sure? It seems the other way around. That is, gems [like jsbundling-rails](https://github.com/rails/jsbundling-rails/blob/7faf59f85303d44f39e78cfce21db09dbe48d60a/lib/tasks/jsbundling/build.rake#L69) hook into `test:prepare`. `rspec-rails` [defines `spec:prepare`](https://github.com/rspec/rspec-rails/blob/8c17b4e5020a4d264e8a79e294c58b5c1ef2b005/lib/rspec/rails/tasks/rspec.rake#L11) (which `jsbundling-rails` also hooks into). Shouldn't the solution be to run `spec:prepare` instead of `test:prepare`?



I know it seems like it shouldn't be the solution... We ran into this on Tarsco, and the `test:prepare` should be what you use. Using what's built in to the core of rails seems like a better solution than throwing pieces from another gem into our `application.rb` file. That's my preference and what we did with Tarsco.

Jeremy-Walton:
@reedrolemodel @Braden-077 Rspec generates code in the rails helper that calls `test:prepare` so my thought was make the default generated code work.

Changing that generated code to use `spec:prepare` does fix the issue though so is probably the better, quick change.

Jeremy-Walton:
@reedrolemodel @Braden-077 So Rspec out of the box does generate that, however, we are overriding that file and generating it from a template anyway so we can just change what our template gives you.

reedrolemodel (approved):

PR #123, opened 2024-08-02
Add ActionCableHelper file and include it in helpers.rb and rspec_generator.rb

dominicmacaulay:
## Why?

With the launch of Turbo 8, capybara was having issues testing features that leverage action cable. This new helper forces capybara to delay its proceeding expectations until after the action cable connection has been established by the page.

## What Changed

* [X] Create action_cable_helper.rb
* [X] Include the helper in helpers.rb
* [X] Add the template reference to rspec_generator.rb

### DIFF
--- a/lib/generators/rolemodel/testing/rspec/rspec_generator.rb
+++ b/lib/generators/rolemodel/testing/rspec/rspec_generator.rb
@@ -26,6 +26,7 @@ module Rolemodel
         template 'support/capybara_drivers.rb', 'spec/support/capybara_drivers.rb'
         template 'support/capybara_testid.rb', 'spec/support/capybara_testid.rb'
         template 'support/helpers/test_element_helper.rb', 'spec/support/helpers/test_element_helper.rb'
+        template 'support/helpers/action_cable_helper.rb', 'spec/support/helpers/action_cable_helper.rb'
         template 'support/helpers.rb', 'spec/support/helpers.rb'
         template 'support/webpacker.rb', 'spec/support/webpacker.rb'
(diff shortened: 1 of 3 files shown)


lib/generators/rolemodel/testing/rspec/rspec_generator.rb
```diff
@@ -26,6 +26,7 @@ def add_spec_files
         template 'support/capybara_drivers.rb', 'spec/support/capybara_drivers.rb'
         template 'support/capybara_testid.rb', 'spec/support/capybara_testid.rb'
         template 'support/helpers/test_element_helper.rb', 'spec/support/helpers/test_element_helper.rb'
+        template 'support/helpers/action_cable_helper.rb', 'spec/support/helpers/action_cable_helper.rb'
```
OutlawAndy, line 29: @Jeremy-Walton any reason the templates directory structure doesn't match the destination structure (missing `spec`)?
Since these files aren't leveraging erb replacement, `copy_file 'spec/support/helpers/action_cable_helper.rb'` would suffice if the paths were the same.  Not a big deal, just curious if there's a reason or just how it's always been..

OutlawAndy (approved):
This looks good, thank you! 👍

lib/generators/rolemodel/testing/rspec/rspec_generator.rb
Jeremy-Walton: Not really sure. I believe it's just how it's been. Probably could be cleaned up in the future.

PR #126, opened 2024-08-14
Update webpack.config.js

OutlawAndy:
## Why?

Assuming a Mailer bundle when one does not exist causes asset compilation failure.

## What Changed

What changed in this PR?

* [x] drop mailer bundle from the default config


OutlawAndy:
@Braden-077 good, call!  I'll do that next. Thanks!

PR #129, opened 2024-08-25
[RMS 2] Tailored Select Generator

Jeremy-Walton:
## Why?

In order to actually use the new tailored select, we need some tooling around it, namely the generator to install it and a simple form input to integrate it into the form builder.

## What Changed

* [X] Add Generator
* [ ] Add tests for generator (currently not working)
* [X] Add simple form custom input
* [X] Add smart select helper for capybara tests
  * [X] Tom Select
  * [X] Tailored Select


lib/rolemodel_rails/version.rb
```diff
@@ -1,3 +1,3 @@
 module RolemodelRails
-  VERSION = '0.9.0'
+  VERSION = '0.9.1'
```
OutlawAndy, line 2: I would bump to 0.10 since this adds a whole new generator.
```suggestion
  VERSION = '0.10.0'
```

spec/generators/rolemodel/rspec_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+require 'generators/rolemodel/testing/rspec/rspec_generator'
+
+RSpec.describe Rolemodel::Testing::RspecGenerator, type: :generator do
+  destination File.expand_path('tmp/', File.dirname(__FILE__))
+
+  before(:all) do
+    prepare_test_app
+    run_generator
```
OutlawAndy, line 9: Try this:

```suggestion
    FileUtils.cd(destination_root) { run_generator }
```

I'm going to work on a PR to move these before/after blocks to the global RSpec config..  Unless you feel like doing that in this PR?  I would support that!

lib/generators/rolemodel/simple_form/templates/app/inputs/tailored_select_input.rb
OutlawAndy, line 1: Is it possible to extract this into a Gem?  I think the generator would still makes since for adding the system test helper and then just `run 'bundle add tailored_select'`

Not a deal breaker, but I think packaging would be really beneficial for something like this.  Which I would expect to evolve & improve over time..

spec/generators/rolemodel/tailored_select_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+require 'generators/rolemodel/tailored_select/tailored_select_generator'
+
+RSpec.describe Rolemodel::TailoredSelectGenerator, type: :generator do
+  destination File.expand_path('tmp/', File.dirname(__FILE__))
+
+  before(:all) do
+    prepare_test_app
+    run_generator
```
OutlawAndy, line 9: try:
```suggestion
    FileUtils.cd(destination_root) { run_generator }
```

Jeremy-Walton: That does it, thanks!

spec/generators/rolemodel/rspec_generator_spec.rb
Jeremy-Walton: I'll hold off on that. Mostly just want to get this generator working

lib/rolemodel_rails/version.rb
Jeremy-Walton: That makes sense. Updated

lib/generators/rolemodel/simple_form/templates/app/inputs/tailored_select_input.rb
Jeremy-Walton: Tailored Select has a few parts.
NPM Package. Can be used on it's own.
Capybara test helper. For when you are using TS with Capybara. This makes sense to be in RMS Rails since that is where our standard config lives.
Custom Simple Form Input. For when you are using Simple form. I think this also makes sense in RMS Rails since again this is where our standard config lives.
Optics Add-on. For when you are using Optics. This one will live in Optics since that is already a separate package and is the one that will define the "bridge" or mapping from one tool to the other.

I don't think a gem makes sense as it would really be a catch all that is highly specific to the tools RMS uses. Unless we build out 1 gem as the "bridge" between Capybara and Tailored Select and another gem as the "bridge" between Simple Form and Tailored Select. Both would be incredibly small so I think leaving things in here makes more sense.

PR #131, opened 2024-10-10
toggle controller inert attribute

Jeremy-Walton:
## Why?

Remote content is still included in the tab index, even if off screen and not scrollable into view. This is relevant when modal or panels are dismissed but still have contents within them. This change allows the [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) data attribute to be toggle along with the active class. This makes the contents non interactive. This makes them both more usable and more accessible.

## What Changed

What changed in this PR?

* [X] Add option to toggle inert attribute when toggling panel and modals
* [X] Add test for the generator

## Screenshots

https://github.com/user-attachments/assets/dc5d068e-7ba2-4e04-b673-93ade9e91153

### DIFF
--- a/lib/generators/rolemodel/modals/templates/app/javascript/controllers/toggle_controller.js
+++ b/lib/generators/rolemodel/modals/templates/app/javascript/controllers/toggle_controller.js
@@ -7,6 +7,10 @@ export default class extends Controller {
     performOnConnect: {
       type: Boolean,
       default: false,
+    },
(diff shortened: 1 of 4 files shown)


wesrich (approved):

OutlawAndy:
@Jeremy-Walton seems like we should always set this, right?  Or is there a use-case where you would not want it?

PR #133, opened 2024-11-21
Update Github pull_request_template.md to include README update, remove Task section

timirwin:
## Why?

Why were the changes needed? What issues were the changes addressing?
(Note: some changes may seem unrelated to the ticket, this is a great place to explain further.)

Linear automatically links to the PR when referenced in the PR name; READMEs are hard to remember to update to confirm they get updated upon PR creation/review

## What Changed

What changed in this PR?

* [x] remove Task reference from pull_request_template generator
* [x] add README update pre-merge checklist to pull_request_template generator
* [x] add/remove from rolemodel_rails pull_request_template.md too
* [x] remove extra checklist items
* [x] add checklist item for updating rolemodel_rails version

## Screenshots

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,3 @@
-## Task
-
-[TR #XXX]()
-
 ## Why?
 
 Why were the changes needed? What issues were the changes addressing?
@@ -14,6 +10,10 @@ What changed in this PR?
 * [ ] Change 1
 * [ ] Change 2
 * [ ] ...
+
+## Pre-merge checklist
+
+* [ ] Update relevant READMEs
 
 ## Screenshots
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,13 +7,12 @@ Why were the changes needed? What issues were the changes addressing?
 
 What changed in this PR?
 
-* [ ] Change 1
-* [ ] Change 2
-* [ ] ...
+* [ ] Change1
 
 ## Pre-merge checklist
 
 * [ ] Update relevant READMEs
+* [ ] Update version number in `lib/rolemodel_rails/version.rb`
 
 ## Screenshots
 
--- a/lib/generators/rolemodel/github/templates/pull_request_template.md
+++ b/lib/generators/rolemodel/github/templates/pull_request_template.md
@@ -1,7 +1,3 @@
-## Task
-
-[TR #XXX]()
-
 ## Why?
 
 Why were the changes needed? What issues were the changes addressing?
@@ -14,6 +10,10 @@ What changed in this PR?
 * [ ] Change 1
 * [ ] Change 2
 * [ ] ...
+
+## Pre-merge checklist
+
+* [ ] Update relevant READMEs
 
 ## Screenshots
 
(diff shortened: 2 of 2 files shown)


reedrolemodel (approved):
I like the Task section removal because it's redundant with Linear integration. Do you add it to the PR title as a prefix? The only advantage there is visibility in git log, which I don't think I ever use.

justwiebe (approved):

timirwin:
> I like the Task section removal because it's redundant with Linear integration. Do you add it to the PR title as a prefix? The only advantage there is visibility in git log, which I don't think I ever use.

I added it as a suffix. Is that good enough, @reedrolemodel?

Jeremy-Walton:
The format that we have typically aimed for is described here. https://github.com/RoleModel/BestPractices/blob/master/git/commit-messages.md

I've always done `[APP#xxx] Task Title (GH#xx)`

lib/generators/rolemodel/github/templates/pull_request_template.md
```diff
@@ -15,6 +11,10 @@ What changed in this PR?
 * [ ] Change 2
 * [ ] ...
```
justwiebe, line 12: Thoughts on removing these extras? I always need to delete them and add my own

reedrolemodel: Hitting Enter after a todo adds a new one automatically, so I agree these are unnecessary.

.github/pull_request_template.md
```diff
@@ -15,6 +11,10 @@ What changed in this PR?
 * [ ] Change 2
 * [ ] ...
```
reedrolemodel, line 12:
```suggestion
```

reedrolemodel:
> > I like the Task section removal because it's redundant with Linear integration. Do you add it to the PR title as a prefix? The only advantage there is visibility in git log, which I don't think I ever use.
> 
> I added it as a suffix. Is that good enough, @reedrolemodel?

My point is I don't need it. The Github PR number is appended by Squash and merge. If I need to look up the task, I can follow the link in my git client to the PR and then follow the link to Linear from there.

reedrolemodel:
> The format that we have typically aimed for is described here. https://github.com/RoleModel/BestPractices/blob/master/git/commit-messages.md
> 
> I've always done `[APP#xxx] Task Title (GH#xx)`

That's how we do it too, except the PR number is appended automatically by Squash merges. For example:
```
556d2507 * [TS#845] Skip SSN presence validation if archived (#1624)
d19bf617 * [TS#841] Fix invitation flow bypass when resetting password (#1623)
114b2a0d * [TS#843] Complete year button (#1621)
```

OutlawAndy (approved):
+1 for removing task list items after `Change 1`.  Otherwise :shipit:

lib/generators/rolemodel/github/templates/pull_request_template.md
OutlawAndy: though, I guess this is the important one.  Isn't it?

.github/pull_request_template.md
OutlawAndy: apply to generator template also

lib/generators/rolemodel/github/templates/pull_request_template.md
reedrolemodel, line 12:
```suggestion
```

wesrich (approved):

PR #135, opened 2024-12-05
Update ESLint config

Braden-077:
## Why?

This PR will allow us to support new apps on ESLint 8+

## What Changed

* [x] Update ESLint config to ESM
* [x] Update ESLint config to the new flat ESLint standard


wesrich (approved):

lib/generators/rolemodel/linters/eslint/templates/eslint.config.js
```diff
@@ -0,0 +1,98 @@
...
+      // No semicolons
+      semi: ['error', 'never'],
+      'no-unexpected-multiline': 'error',
+
+      // 1. We follow the convention of naming "private" properties with a leading underscore
+      // 2. This conflicts with the no-unused-vars config
+      'no-underscore-dangle': 'off',
+
+      // Don't need trailing commas everywhere
+      'comma-dangle': ['error', 'never'],
```
reedrolemodel, line 39: This [is deprecated](https://eslint.org/docs/latest/rules/comma-dangle) but I like dangling commas because they reduce git diff churn.

lib/generators/rolemodel/linters/eslint/templates/eslint.config.js
```diff
@@ -0,0 +1,98 @@
...
+
+      // Don't require naming all non-arrow functions. Modern browsers can usually
+      // detect the correct name if it's assigned to a variable or property.
+      'func-names': 'off',
+
+      // Don't require instance methods to use `this` (the rule name is confusing)
+      'class-methods-use-this': 'off',
+
+      // Allow console.log and similar
+      'no-console': 'off',
```
reedrolemodel, line 54: Can you give an example of needing to commit console.log?

wesrich: > Can you give an example of needing to commit console.log?

That was in the old file; this was just converted syntax/filename.

But on that note, I could see dropping this option.

Braden-077: This was just a migration of one App's ESLint Config. That's likely better to remove.

lib/generators/rolemodel/linters/eslint/templates/eslint.config.js
Braden-077: For reference, it was pulled from an app which had ESLint 8.52, which marks this as valid

reedrolemodel (approved):

PR #139, opened 2025-02-10
Unmount React Components That Get Removed From the DOM

NathanSadler:
## Why?

When a component is mounted using the `react_controller` and the DOM element it is attached to gets removed, the component doesn't actually get unmounted

## What Changed

* [x] Add a `dismount` method to react controller that unmounts react components

## Pre-merge checklist

* [x] Update relevant READMEs
* [ ] Update version number in `lib/rolemodel_rails/version.rb`


Jeremy-Walton (changes requested):
Looks like this method is being deprecated https://18.react.dev/reference/react-dom/unmountComponentAtNode

PR #148, opened 2025-07-30
Add Assign Attributes Override to Soft Destroyable

dominicmacaulay:
## Why?

When updating a soft destroyable record's attributes, it would attempt to save the record when soft destroying it. 
This would become a problem when other attributes, in addition to `_soft_destroy`, were being assigned. If any of those other attributes were invalid, it would fail to soft destroy the record.

This PR fixes that by overriding the `assign_attributes` method, which rails uses under the hood, to ignore all other attributes if `_soft_destroy` is present.

### Example

This is an existing record with a video link attachment. There are validations around the link being real.
WHEN: I edit the link to no longer be valid, then mark the record for soft destruction.
THEN: I attempt to save, but it gives an error saying the record is invalid and cannot be soft destroyed.

https://github.com/user-attachments/assets/24dee7d9-a903-419e-ac22-32798fafbb76

## What Changed

* [x] Override the `assign_attributes` method to ignore other attributes if `_soft_destroy` is passed

## Pre-merge checklist

* ~Update relevant READMEs~
* [x] Update version number in `lib/rolemodel_rails/version.rb`

### DIFF
--- a/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
+++ b/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
@@ -39,6 +39,11 @@ module SoftDestroyable
     end
   end
 
+  def assign_attributes(attrs)
+    super(attrs[:_soft_destroy].present? ? attrs.slice(:_soft_destroy) : attrs)
+  end
+
+
   def soft_destroyed?
     deleted_at.present?
   end
--- a/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
+++ b/lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
@@ -43,7 +43,6 @@ module SoftDestroyable
     super(attrs[:_soft_destroy].present? ? attrs.slice(:_soft_destroy) : attrs)
   end
 
-
   def soft_destroyed?
     deleted_at.present?
   end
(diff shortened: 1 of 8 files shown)


mikehale:
💯

mikehale (approved):

lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable.rb
```diff
@@ -39,6 +39,11 @@ def cascade_soft_destroy(associations)
     end
   end
 
+  def assign_attributes(attrs)
+    super(attrs[:_soft_destroy].present? ? attrs.slice(:_soft_destroy) : attrs)
+  end
+
+
```
wesrich, line 46:
```suggestion
```

lib/generators/rolemodel/soft_destroyable/templates/soft_destroyable_behavior.rb
```diff
@@ -41,6 +41,20 @@
     end
   end
 
+  describe '#assign_attributes' do
+    it 'ignores other attributes when _soft_destroy is present' do
```
wesrich, line 45: Why?
I suppose if one is doing an `_destroy`, they're unlikely to pass in any other attributes (and they'd likely be irrelevant since the record would theoretically be deleted). Is that always the desired thing for `_soft_destroy` though?

PR #157, opened 2025-08-13
Misc updates

OutlawAndy:
## What Changed

What changed in this PR?

* [x] Improve Modal documentation
* [x] simplify flash partial to be a better starting point for new apps
* [x] move flash partial to conventional location
* [x] drop rolemodel_rails version dependency in example apps
* [x] add `respond_to_prompt` spec helper
* [x] improve turbo frame link helpers (add `link_to_top` helper `turbo_frame: "_top"`)
* [x] ensure tests always run against latest example app, without needing to be updated
* [x] improve spec setup, make sure example app is copied to the same location for each run
* [x] update to Ruby 3.4.4

## Pre-merge checklist

* [x] Update relevant READMEs
* [x] Update version number in `lib/rolemodel_rails/version.rb`


wesrich (approved):
One comment, which is just removing a `.DS_Store` file. Other than that, it looks fine to me

PR #186, opened 2026-03-27
Add Release Workflow

OutlawAndy:
## Summary

Version bumps no longer require a manual commit inside every PR. Triggering a workflow now bumps the version, commits, tags, and pushes automatically, and a tag push produces a GitHub Release with generated notes, which in turn publishes the gem to [RubyGems.org](<http://RubyGems.org>) -- three single-purpose workflows instead of one that tried to do everything inline.

## What changed

* `bin/bump_version` **now owns the full git sequence** (add, commit, annotated tag, push) instead of `bump_version.yml` doing it inline in shell. The tag has to point at the version-bump commit, and the only way to guarantee that ordering is to let the same process create both.
* **A** `BUMP_VERSION_AUTOMATED` **guard** blocks the script from running outside the workflow, so a local `bin/bump_version` invocation can't accidentally commit, tag, or push to `master`. This required overriding Thor::Group's `exit_on_failure?` -- it defaults to `false`, so without the override the guard's error would print but the process would still exit `0`.
* `release.yml` **no longer also publishes the gem.** That duplicated the newly-added `publish-package.yml`, which now owns the RubyGems publish on `release: published`. `release.yml` only creates the GitHub Release.
* **The push is atomic.** Code review caught a real bug here: without `--atomic`, a rejected (non-fast-forward) branch update wouldn't stop the new tag from landing anyway, since new tag refs aren't subject to the fast-forward check. That would silently publish a tag -- and trigger a real RubyGems release -- built from a commit that never actually reached `master`. Reproduced the exact cascade in an isolated scratch repo, confirmed `--atomic` closes it.

```mermaid
flowchart TB
  A[workflow_dispatch: bump_version.yml] --> B[bin/bump_version: bump, commit, tag, atomic push]
  B --> C[Tag push triggers release.yml]
  C --> D[GitHub Release created w/ generated notes]
  D --> E[release: published triggers publish-package.yml]
  E --> F[rubygems/release-gem: gem pushed to RubyGems.org]
```

## Testing

* Full `rspec` suite green (88 examples).
* New `spec/bin/bump_version_spec.rb` covers the guard clause: raises without the automation flag, raises on a falsy value, and exits non-zero as a subprocess with the guard message and no file changes.
* `rubocop` clean on touched files (6 pre-existing offenses remain, none introduced by this change).
* Manually verified end-to-end in disposable scratch git repos (not this repo): the full bump -> commit -> tag -> push sequence lands correctly, and separately, that `--atomic` actually blocks the tag from landing when the branch push is rejected.

## Known Residuals

Code review (9 reviewer personas) surfaced three items accepted as-is rather than fixed in this PR:

* `workflow_dispatch` on `bump_version.yml` has no branch/ref restriction -- any collaborator with write access could trigger a bump from a non-`master` branch. Accepted scope boundary.
* `publish-package.yml` triggers on any `release: published` event with no tag-pattern check -- a manually-created GitHub Release would also trigger a real RubyGems publish. This is the trigger this PR was asked to implement; flagging in case a tag-pattern guard is wanted later.
* `bump_version.yml`'s `bump-type` choice input has no explicit `default:`, so "minor by default" may not be what the GitHub UI actually submits when left untouched. Pre-existing YAML that predates this rework.

## Post-Deploy Monitoring & Validation

* **Watch:** the Actions logs for "Bump Version & Publish a GitHub Release", "Create GitHub Release", and "Publish New Gem Version to [RubyGems.org](<http://RubyGems.org>)".
* **Healthy:** `workflow_dispatch` completes green; a `vX.Y.Z` tag and matching commit land on `master`; a GitHub Release appears with generated notes; the new gem version appears on [rubygems.org](<http://rubygems.org>).
* **If it fails:** `bump_version.yml` failing leaves no partial tag/branch state on origin (that's what `--atomic` guarantees) -- safe to just re-trigger. `release.yml` failing after a tag landed can be recovered with `gh release create <tag> --generate-notes`. `publish-package.yml` failing (e.g. RubyGems trusted publishing not yet configured) leaves the GitHub Release published but the gem unpublished -- configure trusted publishing, then re-run the failed job; the underlying `rake release` task detects the tag already exists and skips re-tagging.
* **Window:** the first real release run through this pipeline after merge.

---

## Screenshots

Main (-zsh) — \~CODERoleModelrolemodel_rails 07 16 2026@092637 @2x

---

[Compound Engineering](<https://github.com/EveryInc/compound-engineering-plugin>)

Claude Code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   run_specs:
     name: Rspec
-    runs-on: blacksmith-4vcpu-ubuntu-2404
+    runs-on: ubuntu-latest
     timeout-minutes: 10
     env:
       CI: true
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,0 +1,20 @@
+name: Publish New Gem Version to RubyGems.org
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+        with: { persist-credentials: false }
+      - uses: ruby/setup-ruby@v1
+        with: { bundler-cache: true }
+      - uses: rubygems/release-gem@v1
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,7 +12,10 @@ What changed in this PR?
 ## Pre-merge checklist
 
 * [ ] Update relevant READMEs
-* [ ] Run `bin/bump_version` or `bin/bump_version --patch`
+
+## Notes
+
+Please don't commit version changes within your PR.  Versioning is managed through the Github Workflow.
 
 ## Screenshots
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,7 +15,7 @@ What changed in this PR?
 
 ## Notes
 
-Please don't commit version changes within your PR.  Versioning is managed through the Github Workflow.
+Please don't commit version changes within your PR.  Versioning is managed through the GitHub Workflow.
 
 ## Screenshots
 
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -3,14 +3,13 @@ name: Bump Version & Publish a Github Release
 on:
   workflow_dispatch:
     inputs:
-      version_bump:
-        description: SemVer Bump Type
-        required: true
+      bump-type:
+        description: SemVer Bump Type (minor by default)
+        required: false
         type: choice
         options:
-          - patch
-          - minor
-          - major
+          - --patch
+          - --major
 
 permissions:
   contents: write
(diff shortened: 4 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   run_specs:
     name: Rspec
-    runs-on: blacksmith-4vcpu-ubuntu-2404
+    runs-on: ubuntu-latest
```
wesrich, line 15: This is an interesting change. Any reason to move off blacksmith for CI?

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+        with: { persist-credentials: false }
+      - uses: ruby/setup-ruby@v1
+        with: { bundler-cache: true }
+      - uses: rubygems/release-gem@v1
```
wesrich, line 20: 👍

justwiebe (approved):

wesrich (approved):

PR #189, opened 2026-04-07
CMP-354 | Add some fixes to make the MCP generator more complete

justwiebe:
## Why?

I ran the generator on Compass and it needed some updates

## What Changed

What changed in this PR?

* [x] Add basic policy
* [x] Fix some tests
* [x] Run the doorkeeper migration generator

## Pre-merge checklist


OutlawAndy (commented):
I'm not sure we should assume an Action Policy integration.  Overlapping concerns is a problem we still need to work out a strategy for..  I think, in the interest of greater flexibility it might be better to add a post install message to the user about adding authorization.

justwiebe:
Not needed now that we have RoleModel/rails_mcp

PR #198, opened 2026-07-06
[DEV#97] Update rolemodel rails generators

mark-kraemer:
## Why?

Our generator suite still defaulted new apps to Honeybadger for error tracking and to Yarn 1.22, both of which we've moved away from as a company. Running `rolemodel:all` on a fresh app also pulled in generators (React, SaaS/Devise, GoodJob, etc.) that most new apps don't actually need, making initial setup noisier than it should be and leaving more for the developer to manually undo. This PR brings the generators in line with our current standard stack and gives new apps a leaner, more accurate starting point.

## What Changed

* [x] Replaced Honeybadger with Sentry for both Ruby and JS error reporting, including webpack source map upload and per-user error context
* [x] Added a new `core_setup` generator that runs only the generators appropriate for a brand-new app, separate from the full `all` generator
* [x] Switched projects from Yarn 1.22/global installs to Corepack-managed Yarn 4+, applied consistently everywhere `yarn` commands run
* [x] Added a Heroku deploy-app agent skill and matching GitHub Actions workflows to streamline setting up staging/production deploys
* [x] Pinned the Ruby version in the generated Gemfile so the Heroku buildpack installs the correct version
* [x] Updated the generated production database config to use `DATABASE_URL` so it works out of the box on Heroku
* [x] Made the Tailored Select input generator optional and prompted during Simple Form setup instead of always installed
* [x] Removed the outdated Honeybadger/GoodJob version reminder now that Sentry is the default
* [x] Updated README to document the new core setup workflow and Corepack-based Yarn setup

## Pre-merge checklist

* [x] Update relevant READMEs


lib/generators/rolemodel/heroku/heroku_generator.rb
```diff
@@ -15,6 +15,47 @@ def install_procfile
       template 'Procfile'
     end
 
+    def reference_deploy_app_skill
+      say 'Reference the deploy-app agent skill from rolemodel-rails', :green
```
mark-kraemer: @OutlawAndy - I know we discussed moving this, but it's kinda specific to a fresh build from the generator. We could make it more generic, but the generator just references this skill directly from the AGENTS.md file so that it can be used once and then dev doesn't need any other skill or setup

lib/generators/rolemodel/core_setup/README.md
```diff
@@ -0,0 +1,27 @@
...
+  - [Readme](../readme) — standard project README
+  - [Webpack](../webpack) — Webpack v5 for JS and CSS
+  - [Sentry](../sentry) — error monitoring for Ruby and JavaScript
+  - [Slim](../slim) — Slim templates
+  - [Optics](../optics) — Optics design system
+  - [Testing](../testing) — RSpec, FactoryBot, parallel_tests, TestProf
+    (pass `--js-runner` to include jasmine-playwright-runner)
+  - [SimpleForm](../simple_form) — SimpleForm with our configuration
+  - [Linters](../linters) — Rubocop and ESLint
+  - [UI Components](../ui_components) — flash, modals, navbar
```
OutlawAndy:
```suggestion
  - [UI Components](../ui_components) — flash, the modal pattern, & Turbo 8 support
```

lib/generators/rolemodel/core_setup/USAGE
```diff
@@ -0,0 +1,13 @@
+Description:
+  Runs the core generators every new Rails app should have
+
+  Sets up GitHub config, Heroku deployment, README, Webpack, Sentry, Slim,
+  Optics, testing (RSpec & friends), SimpleForm, linters, UI components,
+  editor config, and Lograge. The result is a new Rails app that is ready
+  to push straight to Heroku.
```
OutlawAndy:
```suggestion
  Sets up GitHub config, Heroku deployment, README, Webpack, Sentry, Slim,
  Optics, testing (RSpec & friends), SimpleForm, linters, Turbo 8+ support,
  and Lograge. The result is an app that is ready to push directly to Heroku.
```

lib/generators/rolemodel/github/templates/workflows/deploy-production.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy to Heroku Production
+
+on:
+  workflow_dispatch:
```
OutlawAndy, line 4: Support for deploying a specific commit or branch name

```suggestion
on:
  workflow_dispatch:
    inputs:
      sha:
        description: 'Specific SHA or Branch name (optional)'
        required: false
        type: string
```

lib/generators/rolemodel/github/templates/workflows/deploy-production.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install Heroku CLI
+        run: |
+          curl https://cli-assets.heroku.com/install.sh | sh
+
+      - uses: akhileshns/heroku-deploy@v3.15.15
+        with:
+          heroku_api_key: ${{ secrets.HEROKU_IT_SUPPORT_API_KEY }}
+          heroku_app_name: ${{ vars.HEROKU_APP_NAME }}
+          heroku_email: ${{ vars.HEROKU_IT_SUPPORT_EMAIL }}
+          healthcheck: '${{ vars.HEROKU_APP_URL }}/up'
```
OutlawAndy, line 34:
```suggestion
          branch: ${{ inputs.sha || 'HEAD' }}
          healthcheck: '${{ vars.HEROKU_APP_URL }}/up'
```

lib/generators/rolemodel/github/templates/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,31 @@
+name: Deploy to Heroku Staging
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
OutlawAndy: support for deploying a specific commit or branch name

```suggestion
  push: { branches: [ staging ] }
  workflow_dispatch:
    inputs:
      sha:
        description: 'Specific SHA or Branch name (optional)'
        required: false
        type: string
```

lib/generators/rolemodel/github/templates/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install Heroku CLI
+        run: |
+          curl https://cli-assets.heroku.com/install.sh | sh
+
+      - uses: akhileshns/heroku-deploy@v3.15.15
+        with:
+          heroku_api_key: ${{ secrets.HEROKU_IT_SUPPORT_API_KEY }}
+          heroku_app_name: ${{ vars.HEROKU_APP_NAME }}
+          heroku_email: ${{ vars.HEROKU_IT_SUPPORT_EMAIL }}
+          healthcheck: '${{ vars.HEROKU_APP_URL }}/up'
```
OutlawAndy:
```suggestion
          branch: ${{ inputs.sha || 'HEAD' }}
          healthcheck: "${{ vars.HEROKU_APP_URL }}/up"
```

lib/generators/rolemodel/sentry/sentry_generator.rb
```diff
@@ -0,0 +1,116 @@
...
+      say 'Setting up Sentry for Ruby error reporting', :green
+
+      copy_file 'config/initializers/sentry.rb'
+    end
+
+    def add_user_context
+      say 'Adding Sentry user context to ApplicationController', :green
+
+      inject_into_class 'app/controllers/application_controller.rb', 'ApplicationController',
+                        "  before_action :set_sentry_user, if: :user_signed_in?\n"
```
OutlawAndy: `user_signed_in?` is from the authentication generators isn't it? Seems like in that case `Current.user=` could also set the sentry user..

lib/generators/rolemodel/sentry/sentry_generator.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      inject_into_file 'webpack.config.js', after: "'process.env.RAILS_ENV': JSON.stringify(process.env.RAILS_ENV),\n" do
+        <<-JS
+      'process.env.SENTRY_DSN': JSON.stringify(process.env.SENTRY_DSN),
+      'process.env.SENTRY_ENVIRONMENT': JSON.stringify(process.env.SENTRY_ENVIRONMENT),
+      'process.env.SENTRY_AUTH_TOKEN': JSON.stringify(process.env.SENTRY_AUTH_TOKEN),
+        JS
+      end
+
+      gsub_file 'webpack.config.js', "    })\n  ].filter(Boolean)", <<-JS.chomp
```
OutlawAndy, line 93: I've been looking at this line for a while now.  I feel like I understand a little less about programming every time I read it..

lib/generators/rolemodel/sentry/sentry_generator.rb
```diff
@@ -0,0 +1,116 @@
...
+      !process.env.CI &&
+      sentryWebpackPlugin({
+        authToken: process.env.SENTRY_AUTH_TOKEN,
+        org: 'rolemodel-software',
+        project: '#{app_name}',
+        telemetry: false,
+        applicationKey: 'app-frontend'
+      })
+  ].filter(Boolean)
+      JS
```
OutlawAndy, line 107: How come this is such a complicated process?  I feel like it shouldn't be so hard to add sentry to an app 😭

lib/generators/rolemodel/sentry/USAGE
```diff
@@ -0,0 +1,13 @@
...
+Example:
+    rails generate rolemodel:sentry
+
+    This will:
+        Add the sentry-rails gem
+        Create config/initializers/sentry.rb
+        Add Sentry user context to app/controllers/application_controller.rb
+        Add @sentry/browser and @sentry/webpack-plugin JS dependencies
+        Create app/javascript/initializers/sentry.js
+        Wire the Sentry webpack plugin into webpack.config.js
```
OutlawAndy, line 13: is that all...  Feels like more 😮‍💨

lib/generators/rolemodel/simple_form/simple_form_generator.rb
```diff
@@ -16,5 +16,11 @@ def add_files
       copy_file 'config/initializers/simple_form.rb'
       copy_file 'config/locales/simple_form.en.yml'
     end
+
+    def add_tailored_select_input
+      return unless yes?('Would you like to include the tailored_select custom input?')
```
OutlawAndy: also needs a class option like:

```ruby
  class_option :tailored_select, type: :boolean, default: false, desc: "Install the tailored_select experimental component input"
```

```suggestion
      return unless options.tailored_select?
```

Though, I would actually put the template file back where it was and just have this method conditionally delete it like:

```ruby
  remove_file 'app/inputs/tailored_select_input.rb' unless options.tailored_select?
```

lib/generators/rolemodel/simple_form/USAGE
```diff
@@ -9,3 +9,7 @@ Example:
         config/locales/simple_form.en.yml
         simple_form input wrapper classes
         simple_form template file
+
+    All standard custom inputs are included by default. You will be
+    prompted to confirm whether to include the tailored_select custom
+    input, since not every project uses the tailored-select web component.
```
OutlawAndy: I don't think this is necessary.  Tailored Select is not fully baked, and adding the class option will provide the visibility in the usage output already.

```suggestion
```

lib/generators/rolemodel/tailored_select/tailored_select_generator.rb
```diff
@@ -7,6 +7,7 @@ class TailoredSelectGenerator < GeneratorBase
     def add_tailored_select_package
       say 'Installing Tailored Select package', :green
 
+      ensure_yarn
```
OutlawAndy, line 10: what does `ensure_yarn` do?

Also, the cross-contamination between generators is giving me indigestion..  Not your fault, of course.  I didn't even know this generator existed, but since it does maybe we should move the input template here..  Wonder if there is way to ask if the app is using simple form from this generator??

lib/generators/rolemodel/webpack/templates/webpack.config.js
```diff
@@ -1,7 +1,6 @@
 import path from 'path'
 import webpack from 'webpack'
 import TerserPlugin from 'terser-webpack-plugin'
-import HoneybadgerSourceMapPlugin from '@honeybadger-io/webpack'
```
OutlawAndy, line 4: Okay, Know I understand where some of the complexity in the Sentry generator is coming from.  For HoneyBadger, the code just lived in this file..  Not sure which approach I hate least...  I need a vacation.

lib/generators/rolemodel/webpack/webpack_generator.rb
```diff
@@ -3,10 +3,9 @@ class WebpackGenerator < GeneratorBase
     source_root File.expand_path('templates', __dir__)
 
     DEV_DEPS = %w[
-      @honeybadger-io/webpack
-      @honeybadger-io/js
       esbuild
       esbuild-loader
+      terser-webpack-plugin
```
OutlawAndy: I feel like the latest version of webpack just removed this plugin..  there are several broken dependabot updates laying around our various repos..

lib/generators/rolemodel/webpack/webpack_generator.rb
```diff
@@ -50,15 +55,6 @@ def add_npm_packages
...
 
-    def honeybadger_setup
-      say 'Setting up Honeybadger for JS error reporting', :green
-
-      copy_file 'app/javascript/initializers/honeybadger.js'
-      append_to_file 'app/javascript/application.js', <<~JS
-        import './initializers/honeybadger'
-      JS
-    end
-
```
OutlawAndy, line 61: I wonder if we could use `class_options` to coordinate overlapping concerns. Like if `sentry` do the webpack related sentry crap here??

lib/generators/rolemodel/heroku/templates/deploy-app/SKILL.md
OutlawAndy, line 1: lib is inserted into the load_path of consuming apps, so I don't think this should go here...

I don't know where to put it, but especially if its only relevant for new apps, I don't think it should permanently reside inside every RoleModel application..

What if we just put it in Google Drive and have a shell script that downloads and dumps it to standard out?  Then you can tell Calude/etc.  Run this for the context you need...

lib/rolemodel/generator_base.rb
```diff
@@ -13,5 +13,42 @@ class GeneratorBase < ::Rails::Generators::Base
     def run_bundle
       bundle_command("install --quiet", "BUNDLE_IGNORE_MESSAGES" => "1")
     end
+
+    # Enable Corepack and pin the project to Yarn 4+ (instead of the classic
+    # Yarn 1.22). Idempotent, so any generator can call it before running a
+    # `yarn` command to guarantee the modern toolchain is in place.
+    def ensure_yarn
```
OutlawAndy, line 20: not in love, but you've already worn me down by sheer line count. If you need me I'll be in the corner, crying.

spec/generators/rolemodel/core_setup_generator_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      ['rolemodel:slim'],
+      ['rolemodel:optics:all'],
+      ['rolemodel:testing:all'],
+      ['rolemodel:simple_form'],
+      ['rolemodel:linters:all'],
+      ['rolemodel:ui_components:flash'],
+      ['rolemodel:ui_components:modals'],
+      ['rolemodel:lograge']
+    ]
+  end
```
OutlawAndy, line 28: bad Claude, stop that!


```suggestion
```

OutlawAndy (commented):
@mark-kraemer lets sync on this tomorrow

lib/generators/rolemodel/heroku/heroku_generator.rb
OutlawAndy: I think we should put it somewhere external, don't care where, and then pass to Claude/Copilot via download URL.  It's not a skill anyone needs installed on their machine, but it also doesn't belong on the load path of every RoleModel project for ever.

lib/rolemodel/generator_base.rb
```diff
@@ -13,5 +13,86 @@ class GeneratorBase < ::Rails::Generators::Base
...
+      return if File.read(gitignore).include?('/.yarn/install-state.gz')
+
+      append_to_file '.gitignore', "\n/.yarn/install-state.gz\n"
+    end
+
+    # Single source of truth for wiring the Sentry webpack plugin into
+    # webpack.config.js. Shared by the webpack generator (which owns the config
+    # file) and the sentry generator (for standalone runs). Idempotent and a
+    # no-op if there's no webpack.config.js or it's already wired.
+    def wire_sentry_into_webpack
```
mark-kraemer: Remove from here and add somewhere else

lib/generators/rolemodel/heroku/heroku_generator.rb
mark-kraemer: Move into the template directory and output when run so a user can copy the file path to use in their LLM

OutlawAndy (approved):
discussed with Mark

lib/generators/rolemodel/heroku/heroku_generator.rb
```diff
@@ -15,6 +15,33 @@ def install_procfile
...
+      # none of which exist on Heroku, where the Postgres add-on provides a full DATABASE_URL.
+      # Replace the whole block with a url-based config.
+      gsub_file 'config/database.yml',
+                /^production:\n(?:[ \t]+.*\n?)+/,
+                <<~YAML
+                  production:
+                    <<: *default
+                    url: <%= ENV["DATABASE_URL"] %>
+                YAML
+    end
```
OutlawAndy, line 43: I missed this before it merged, but this is not necessary. Rails uses `ENV['DATABASE_URL']` if it exists, regardless of what config/database.yml says.

PR #199, opened 2026-07-06
Document Rolemodel::Utility::TaskTools

OutlawAndy:
And add additional helper methods

## Why?

Add additional rake task helpers and add documentation.

## What Changed

* [x] Add documentation for Rolemodel::Utility::TaskTools mixin.
* [x] Add additional task helper methods to the module.

## Pre-merge checklist

* [x] Update relevant READMEs

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,117 @@ bin/rails g
 * [Tailored Select](./lib/generators/rolemodel/tailored_select)
 * [Lograge](./lib/generators/rolemodel/lograge)
 
+## Utilities
+
+### `Rolemodel::Utility::TaskTools`
+
+A mixin of helper methods for writing friendlier, more informative Rake tasks. It provides consistent, migration-style console output, progress indication for long-running loops, an opt-in dry-run mode, and sanitized positional task arguments.
+
+#### Requiring & Including
+
+The module is available at runtime (rolemodel-rails must not be restricted to the `development` gem group). Require it and `include` it inside the `namespace` block of your `.rake` file:
+
+```ruby
+require 'rolemodel/utility/task_tools'
+
+namespace :reports do
+  include Rolemodel::Utility::TaskTools
+
+  task total: :environment do
+    @total = GeneratedReport.count
+  end
+
+  namespace :clear do
+    desc 'Delete all non-current report records'
+    task expired: :total do
+      say_with_time "Detecting & Deleting Expired Reports among #{@total} Total" do
+        deleted_reports = 0
+        GeneratedReport.find_each.with_index do |report, index|
+          indicate_progress(index, @total)
+          next if report.current?
+
+          report.destroy unless dry_run?
+          deleted_reports += 1
+        end
+        say "#{deleted_reports}/#{@total} Records Deleted"
+      end
+    end
+  end
+end
+```
+
+> [!NOTE]
+> `include` inside a `namespace` block adds the helper methods to the anonymous object that evaluates your task bodies, making them (and any plain methods you define alongside them) available to every task in that namespace.
+
+#### Helper Methods
+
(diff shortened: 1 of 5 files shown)


lib/rolemodel/utility/task_tools.rb
```diff
@@ -32,11 +33,37 @@ def indicate_progress(index, total = nil, report_interval: 9)
         print("#{PROGRESS.next} #{to_percent(index, total) if total}\r") # rubocop:disable Rails/Output
       end
 
+      ##
+      # Enables a pattern for dry-run mode in tasks. Assumes usage of other utility methods
+      # like `say_with_time` and `indicate_progress` that provide user feedback, separate from the actions they perform.
+      #
+      # Tasks can be updated to support this mode by adding `unless dry_run?` to lines of code that write to the file system or database.
+      def dry_run?
+        ENV['DRY_RUN'].present?
```
wesrich, line 42: Should we check like, active support Boolean cast, to handle "false" being passed in? Sure, it's a little weird, but might be a nicer option.

README.md
```diff
@@ -102,6 +102,117 @@ bin/rails g
...
+##### `say_with_time(message, &block)`
+
+Wraps `say` and prints the block's real execution time as an indented subitem. Use it to announce and time a unit of work.
+
+```ruby
+say_with_time "Deleting all #{@total} Records" do
+  GeneratedReport.destroy_all unless dry_run?
+end
+#=> -- Deleting all 42 Records
+#=>    -> 0.0198s
```
wesrich, line 169: Been a while, but I think in a migration, 'say_with_time' when the last argument returns numeric, will tell that as a subitem too. Maybe we could (eventually?) add support for that?

wesrich (approved):

lib/rolemodel/utility/task_tools.rb
OutlawAndy: my thought was that if it exists than it's on.  Even if you were running multiple tasks, I can't picture wanting some of them dry_run? and some of them not..

Is there a specific use-case you had in mind?

README.md
OutlawAndy: I thought about that, the code is based on the Rails implementation. I left that part out because in a migration you can assume your dealing with Records, so the Rails migration helper is literally `say "#{result} records", :subitem`. I'm not sure the use-cases for this method always similar to that one, but I could be wrong..

Tell you what.  Start using these helpers in your apps and spread the word.  Then we'll have a sample size larger than 1 (me) 😆

lib/rolemodel/utility/task_tools.rb
wesrich: > my thought was that if it exists than it's on. Even if you were running multiple tasks, I can't picture wanting some of them dry_run? and some of them not..
> 
> Is there a specific use-case you had in mind?

It's a reasonable first pass. Just thinking if someone passed in `false` for some reason, they might be surprised. But then, why would they do that? 🤷‍♂️ Not a high enough impact to be a blocker.

README.md
wesrich: > I thought about that, the code is based on the Rails implementation. I left that part out because in a migration you can assume your dealing with Records, so the Rails migration helper is literally `say "#{result} records", :subitem`. I'm not sure the use-cases for this method always similar to that one, but I could be wrong..
> 
> Tell you what. Start using these helpers in your apps and spread the word. Then we'll have a sample size larger than 1 (me) 😆

Totally fair. I'm not sure that logging the number of records is really all that helpful, since # of records may not be what's actually important here. (And, interestingly enough, I feel like a lot of the time getting the return value of the block to be # of records-compatible in the first place is more work than is useful.)

PR #201, opened 2026-07-21
Turbo generators

OutlawAndy:
## Why?

Extract the ui_components:modals generator into multiple turbo generators that can be run separately for subsets of the behavior.

## What Changed

What changed in this PR?

* [x] add rolemodel:turbo:ready
* [x] add rolemodel:turbo:confirm
* [x] add rolemodel:turbo:form
* [x] add rolemodel:turbo:modals [--panels]
* [x] add rolemodel:turbo:all
* [x] remove rolemodel:ui_components:modals
* [x] update rolemodel:core_setup
* [x] extract command: rolemodel:yarn:setup
* [x] update spec support
* [x] update rolemodel:sentry (extract controller concern)
* [x] remove SCSS references
* [x] add bin/update_dep_versions [--node VERSION] [--ruby VERSION] script

## Pre-merge checklist

* [x] Update relevant READMEs


lib/generators/rolemodel/core_setup/core_setup_generator.rb
```diff
@@ -1,18 +1,18 @@
 module Rolemodel
   class CoreSetupGenerator < ::Rolemodel::GeneratorBase
     def run_the_core_generators
+      generate 'rolemodel:slim'
```
Jeremy-Walton, line 6: Worth checking that the USAGE.md files are all up to date

lib/generators/rolemodel/optics/base/base_generator.rb
```diff
@@ -13,9 +13,9 @@ def add_optics_package
       def copy_templates
         say 'importing stylesheet', :green
 
-        prepend_to_file Dir.glob('app/assets/stylesheets/application.*').first, <<~SCSS
+        prepend_to_file application_stylesheet_path, <<~CSS
           @import '@rolemodel/optics/dist/css/optics';
```
Jeremy-Walton, line 17: The icon helper generator could swap this line out for the specific icon pack version. See https://docs.optics.rolemodel.design/?path=/docs/overview-addons--docs#icon-fonts

lib/generators/rolemodel/saas/devise/devise_generator.rb
```diff
@@ -161,11 +161,11 @@ def add_login_styles
         say 'importing login stylesheet', :green
         copy_file 'app/assets/stylesheets/login.css'
```
Jeremy-Walton, line 162: Per discussion with Andy. This generator probably needs to get rewritten.

* New Rails 8 authentication.
* Simplify the models being created (user, org, etc)

lib/generators/rolemodel/saas/devise/devise_generator.rb
```diff
@@ -161,11 +161,11 @@ def add_login_styles
...
 
-        return unless File.exist?(File.join(destination_root, 'app/assets/stylesheets/application.scss'))
+        return unless File.exist?(application_stylesheet_path)
 
-        append_to_file 'app/assets/stylesheets/application.scss', <<~SCSS
+        append_to_file application_stylesheet_path, <<~CSS
           @import 'login';
```
Jeremy-Walton, line 167: Probably a follow up but I intend to simplify the login pattern. There are better ways to write this in a cleaner way.

lib/generators/rolemodel/saas/devise/devise_generator.rb
Jeremy-Walton: Omni auth, simplified devise generator?

lib/generators/rolemodel/saas/devise/devise_generator.rb
Jeremy-Walton: https://github.com/RoleModel/html_css_card_game_gofish/blob/main/index.html
I think just 

```
.op-page {
  .op-page__main {
    .op-page__main-content.op-page__main-content--centered {
      block-size: 100dvh;
      place-content: center;
      padding-block: 0;
    }
  }
}
```

spec/generators/rolemodel/simple_form_generator_spec.rb
```diff
@@ -1,12 +1,12 @@
 RSpec.describe Rolemodel::SimpleFormGenerator, type: :generator do
```
Jeremy-Walton, line 1: Would be worth investigating this pattern Justin made for applying the right optics classes without having to reach into input or wrapper_html to set classes

in `config/initializers/simple_form.rb`
```ruby
module OpticsFormGroupConfig
  def html_classes(input, options)
    css = super(input, options)
    css << 'form-group--inline' if options[:inline]
    css << 'form-group--compact' if options[:compact]
    css
  end
end

module OpticsFormControlConfig
  def additional_classes
    @additional_classes ||= begin
      classes = super
      classes << 'form-control--no-border' if options[:no_border]
      classes << "form-control--#{options[:size]}" if options[:size]
      classes
    end
  end
end

Rails.application.config.to_prepare do
  SimpleForm::Wrappers::Root.prepend OpticsFormGroupConfig
  SimpleForm::Inputs::Base.prepend OpticsFormControlConfig
end
```

Could also look at the components (like b.use ...) as a potentially cleaner way to achieve this.

lib/generators/rolemodel/turbo/modals/modals_generator.rb
```diff
@@ -48,18 +36,15 @@ def to_panel_or_not_to_panel
...
           end
 
-          inject_into_file 'app/assets/stylesheets/application.scss',
-                          after: "@import '@rolemodel/optics/dist/css/optics';\n" do
-            <<~SCSS
-              @import '@rolemodel/optics/dist/css/addons/panel';
-            SCSS
-          end
+          append_to_file application_stylesheet_path, <<~CSS
+            @import '@rolemodel/optics/dist/css/addons/panel';
```
Jeremy-Walton, line 46: I think panel can go away. Modal became a first class citizen a while ago but panel isn't really in a clean state from Optics.

Probably worth rewriting into a drawer pattern as a first class citizen in Optics and rebuilding this.

lib/generators/rolemodel/turbo/modals/modals_generator.rb
```diff
@@ -39,7 +28,6 @@ def inject_into_layout
         inject_into_file 'app/views/layouts/application.html.slim', after: /\bbody.*\n/ do
           optimize_indentation <<~SLIM, 4
             = turbo_frame_tag 'modal'
```
Jeremy-Walton, line 30: Would be good to convert the modal pattern to use native dialogs.

Also rename the toggle controller.

`top_layer_controller.js`
```js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  static targets = ["form"]
  static values = {
    openOnConnect: {
      type: Boolean,
      default: false
    },
    openOnError: {
      type: Boolean,
      default: false
    }
  }

  connect() {
    this.scrollLockClass = "op-scroll-lock"

    if (this.openOnConnectValue) this.open()
    if (this.openOnErrorValue) this.openWithoutAnimation()

    if (this.hasFormTarget) {
      this.formTarget.addEventListener("turbo:submit-end", this.handleSubmitEnd.bind(this))
    }
  }

  disconnect() {
    if (this.hasFormTarget) {
      this.formTarget.removeEventListener(
        "turbo:submit-end",
        this.handleSubmitEnd.bind(this)
      )
    }
  }

  open() {
    // next-tick to allow animation-in after connect
    setTimeout(() => {
      this.element.showModal()
      this.#lockBodyScrolling(this)
    }, 0)
  }

  openWithoutAnimation() {
    this.element.showModal()
    this.#lockBodyScrolling(this)
  }

  close() {
    this.element.requestClose()
    this.#unlockBodyScrolling(this)
  }

  submit(event) {
    event.preventDefault()
    event.target.disabled = true

    if (this.hasFormTarget) {
      this.formTarget.requestSubmit()
    }
  }

  handleSubmitEnd(event) {
    const { success, fetchResponse } = event.detail

    // Close the dialog only if the submission was successful
    // and the response doesn't indicate an unprocessable entity
    if (success && fetchResponse?.response?.ok) {
      this.close()
    }
  }

  #lockBodyScrolling() {
    document.documentElement.classList.add(this.scrollLockClass)
  }

  #unlockBodyScrolling() {
    document.documentElement.classList.remove(this.scrollLockClass)
  }
}
```

This is from my side project where I called it sheet, but could rename to modal for this context.

`sheet.html.slim`
```slim
doctype html
html data-theme-mode='dark'
  head
    = render 'layouts/head'
  body
    = render "alerts"
    = render "confirm"
    = turbo_frame_tag 'sheet' do
      // Apply a class with:
      // = content_for(:sheet_additional_classes) do
      //   | sheet--half
      dialog(
        class=class_names(\
          'sheet',\
          "#{yield :sheet_additional_classes}",\
          'sheet-disable-animation': response.message.starts_with?('Unprocessable')\
        )
        data-controller="top-layer"
        data-top-layer-open-on-connect-value=(response.message == 'OK')
        data-top-layer-open-on-error-value=(response.message.starts_with?('Unprocessable'))
        data-turbo-temporary="true"
      )
        header.sheet__header
          button type='submit' class="btn btn--icon" data-action="top-layer#close"
            = icon('x')
            span.sr-only Close
          = yield :sheet_title
          = yield :sheet_action

        .sheet__body
          = yield

```

Jeremy-Walton: Also the css. again, sheet specific but the starting style part is important and Optics modal already uses starting style.

```css
  .sheet {
    --_co-sheet-animation-duration: 200ms;
    --_co-sheet-radius: 35px;
    --_co-sheet-spacing: var(--op-space-small) var(--op-space-x-large);

    top: auto;
    max-block-size: 100%;
    max-inline-size: 100%;
    block-size: 100%;
    inline-size: 100%;

    border: none;
    padding: 0;

    border-start-start-radius: var(--_co-sheet-radius);
    border-start-end-radius: var(--_co-sheet-radius);

    background-color: var(--op-color-neutral-plus-seven);
    color: var(--op-color-neutral-on-plus-seven);

    transform: translateY(100%);
    transition:
      display var(--_co-sheet-animation-duration) allow-discrete,
      overlay var(--_co-sheet-animation-duration) allow-discrete,
      transform var(--_co-sheet-animation-duration);

    &[open] {
      transform: translateY(0);
    }

    &.sheet--half {
      block-size: 50%;
      max-block-size: 50%;
    }

    &.sheet--three-quarter {
      block-size: 75%;
      max-block-size: 75%;
    }

    .sheet__header {
      position: sticky;
      inset-block-start: 0;
      inset-inline: 0;

      display: grid;
      grid-template-columns: 5rem 1fr 5rem;
      align-items: center;
      padding: var(--_co-sheet-spacing);
      background-color: var(--op-color-neutral-plus-seven);
      color: var(--op-color-neutral-on-plus-seven);
      z-index: var(--op-z-index-header);

      :nth-child(3) {
        margin-inline-start: auto;
      }
    }

    .sheet__title {
      margin: 0;
      font-weight: var(--op-font-weight-medium);
      font-size: var(--op-font-large);
      text-align: center;
    }

    .sheet__body {
      padding: var(--_co-sheet-spacing);
    }
  }

  @starting-style {
    .sheet:not(.sheet-disable-animation)[open] {
      transform: translateY(100%);
    }
  }
```

lib/generators/rolemodel/ui_components/navbar/navbar_generator.rb
```diff
@@ -17,7 +17,7 @@ def insert_navbar_before_content
         insert_into_file 'app/views/layouts/application.html.slim', before: /\s+\.app__content/ do
           optimize_indentation <<~SLIM, 4
 
-            = render 'layouts/navbar'
```
Jeremy-Walton, line 20: Worth investigating. What is the navbar being added? Is this necessary? Why is shoelace being added and what is being used for in this context

lib/generators/rolemodel/webpack/templates/postcss.config.cjs
```diff
@@ -1,9 +1,6 @@
 module.exports = {
   syntax: require('postcss-scss'),
```
Jeremy-Walton, line 2: This may not be needed anymore

lib/generators/rolemodel/webpack/templates/webpack.config.js
Jeremy-Walton, line 57: Can probably simplify this test. See the comment I added in slack the other day for a simpler test regex

Jeremy-Walton (approved):

lib/generators/rolemodel/turbo/modals/modals_generator.rb
Jeremy-Walton:
```css
  .op-scroll-lock {
    overflow: hidden;
  }
  ```

